PR #2, opened 2018-09-04
Openstack Exporter - Task #14702

jakecoll:
Adds an ansible playbook for extending prometheus monitoring to track metrics from openstack exporter.


inventories/chi_uc/group_vars/openstack_exporter
```diff
@@ -0,0 +1,11 @@
+---
+openstack_prometheus_exporter_enable: yes
```
diurnalist: I think we can get rid of this variable, since including the role == enabling the exporter. It's a bit redundant to keep it for this case.

inventories/chi_uc_dev/group_vars/openstack_exporter
```diff
@@ -0,0 +1,12 @@
+---
+openstack_prometheus_exporter_enable: yes
```
diurnalist: Same as `chi_uc`

inventories/chi_uc_dev/hosts
```diff
@@ -15,9 +15,12 @@ admin02
 # The Ceph cluster monitoring host (usually the master)
 # [ceph_exporter]
 
+<<<<<<< HEAD
```
diurnalist: Looks like a merge conflict here, please resolve before proceeding.

roles/openstack_exporter/defaults/main.yml
```diff
@@ -0,0 +1,16 @@
+---
+openstack_prometheus_exporter_port: 19103
+openstack_prometheus_exporter_enable: False
```
diurnalist: Can remove this pre the prior comments

roles/openstack_exporter/defaults/main.yml
```diff
@@ -0,0 +1,16 @@
...
+openstack_prometheus_exporter_docker_image: rakeshpatnaik/prometheus-openstack-exporter:v0.2
+openstack_prometheus_exporter_service_name: prometheus-openstack-exporter
+openstack_prometheus_exporter_config_path: /etc/prometheus/openstack-exporter
+openstack_prometheus_exporter_project_name: 'openstack'
```
diurnalist: You don't need quotes here or really for any of the strings; I would remove them for consistency.

roles/openstack_exporter/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+        dest: "{{ openstack_prometheus_exporter_config_path }}/env-list"
+      notify:
+        - restart openstack exporter
+    - name: Configure systemd service.
+      template:
+        src: prometheus_exporter.service.j2
+        dest: "/etc/systemd/system/{{ openstack_prometheus_exporter_service_name }}.service"
+      notify:
+        - restart openstack exporter
+
```
diurnalist: I think you may need to remove this newline, the newline should go after the 'tags' section instead. Not sure how graceful YAML is at parsing this, but it's better to be safe than sorry.

roles/openstack_exporter/templates/prometheus_exporter.env-list.j2
```diff
@@ -0,0 +1,23 @@
+[client]
```
diurnalist: This is not valid in a Docker env file, please remove. 

> You can also load the environment variables from a file. This file should use the syntax <variable>=value (which sets the variable to the given value) or <variable> (which takes the value from the local environment), and # for comments.
[source](https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file)

roles/openstack_exporter/templates/prometheus_exporter.service.j2
```diff
@@ -0,0 +1,17 @@
+# {{ ansible_managed }}
```
diurnalist, line 1: FYI, what this does is it adds a header when Ansible writes the file, so an operator that comes later can know that the file was automatically generated (and thus shouldn't be modified directly, as any changes would be overwritten.)

roles/openstack_exporter/templates/prometheus_exporter.service.j2
```diff
@@ -0,0 +1,17 @@
...
+[Unit]
+Description=OpenStack metrics exporter
+After=docker.service
+
+[Service]
+Restart=always
+ExecStart=/bin/docker run --name=%N \
```
diurnalist, line 8: FYI `%N` means "the name of the systemd service"

roles/openstack_exporter/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+- name: Configure OpenStack exporter.
+  block:
+    - name: Create config directory.
+      file:
+        path: "{{ openstack_prometheus_exporter_config_path }}"
+        state: directory
+    - name: Create OpenStack env variable list.
+      template:
+        src: prometheus_exporter.env-list.j2
+        dest: "{{ openstack_prometheus_exporter_config_path }}/env-list"
```
diurnalist: This is kind of a funny name for this file, I usually do like `{{ *_service_name }}.env` to make it obvious what it is and which service is using it. Can we try to follow that convention?

roles/openstack_exporter/templates/prometheus_exporter.service.j2
```diff
@@ -0,0 +1,17 @@
...
+                          --net=host \
```
diurnalist: I wouldn't use `net=host` for this, rather expose a port explicitly using `--publish`. See [here](https://docs.docker.com/engine/reference/run/#expose-incoming-ports) for more about what this means. Basically we would publish a port (configurable in the role) that binds to whatever port the container uses by default, looks like that is [9103](https://github.com/att-comdev/prometheus-openstack-exporter/blob/master/Dockerfile#L13) in this case. Something like this can accomplish that:

```jinja
--publish {{ openstack_prometheus_exporter_port }}:9103
```

We should avoid `net=host` if we can, I think, as it means the container shares all the same network interfaces as the host OS, which can mean that the container is allowed to talk to more things than it should.

diurnalist (changes requested):
A few notes, mostly about port configuration and how the env file is set up. Also need to watch out for bad Git merges committed. Otherwise, this looks great! 👍

roles/openstack_exporter/tasks/main.yml
jakecoll: yeppers. It's changed.

roles/openstack_exporter/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+- name: Configure OpenStack exporter.
+  block:
+    - name: Create config directory.
+      file:
+        path: "{{ openstack_prometheus_exporter_config_path }}"
+        state: directory
+    - name: Create OpenStack env variable list.
+      template:
+        src: prometheus_exporter.env.list.j2
+        dest: "{{ openstack_prometheus_exporter_config_path }}/{{ openstack_prometheus_exporter_service_name }}.env"
```
diurnalist: Can you add a `mode: "0600"` to this and test that it still works? Since this file has secrets in it, it is always best to ensure it can't be read by every user.

diurnalist (approved):
One final comment but I'm giving a prelim 👍 anyways. Nice work

roles/openstack_exporter/tasks/main.yml
jakecoll: yep, it works.

PR #3, opened 2018-09-10
Update prometheus pb

jakecoll:
Added openstack_exporter to prometheus playbook.


diurnalist (approved):
Looks good to me! Why don't you also update the inventory for UC dev/prod and then we can merge this

PR #4, opened 2018-09-14
Simplify prometheus config generation

diurnalist:
Currently we have a lot of boilerplate when setting up prometheus static
configs. This simplifies that a bunch by pulling the internal addresses
out to a fact first, then creating a 'jobs' structure that associates a
scrape job with an inventory group. The template references ansible
variables directly instead of relying on some magic vars being placed on
the template context.


diurnalist:
cc @jakecoll this fixes some of the issues we're currently seeing w/ scraping.

PR #5, opened 2018-09-17
Added extended prometheus-openstack-exporter image to playbooks.

jakecoll:
This just replaces the public docker image we were using with our own image, and modifies one environment variable.

Extended OpenStackExporter:
https://github.com/ChameleonCloud/prometheus-openstack-exporter


diurnalist (approved):
Nice, simple change 👍

PR #6, opened 2018-10-30
Alertmanager

jakecoll:
Updates prometheus to expose and connect to alert manager. 

Note: There doesn't seem to be a convenient way to send a dummy alert to the alertmanager, but I'm not seeing any errors after enabling "alerting" in the promtheus playbook.


inventories/chi_uc/group_vars/prometheus
```diff
@@ -0,0 +1,13 @@
+---
+prometheus_alertmanager_hostname: alert.uc.chameleoncloud.org
+
+slack_api_url: !vault |
```
diurnalist: This is already available in the `group_vars/all` file, which applies to all hosts. So you don't need it here as well.

playbooks/jupyterhub.yml
```diff
@@ -1,4 +1,4 @@
----
+R---
```
diurnalist: Accidental `R` in here

inventories/chi_uc_dev/group_vars/prometheus
```diff
@@ -0,0 +1,13 @@
+---
+prometheus_alertmanager_hostname: alert.dev.uc.chameleoncloud.org
+
+slack_api_url: !vault |
```
diurnalist: This can go to `group_vars/all` - the reason for this is the Slack API url may be re-used for other things (I think).

playbooks/prometheus.yml
```diff
@@ -24,3 +32,48 @@
...
+                AuthType Basic
+                AuthName "Alertmanager authentication"
+                AuthBasicProvider socache external
+                AuthExternal keystone-user
+                AuthExternalProvideCache On
+                AuthnCacheProvideFor keystone-user
+
+                Require user {{ prometheus_alertmanager_users | join(' ') }}
+
+                # preserve Host header to avoid cross-origin problems
```
diurnalist, line 72: You might not need this; this could be a JupyterHub thing

roles/apache/tasks/main.yml
```diff
@@ -20,6 +20,7 @@
   template:
     src: "{{ item }}.j2"
     dest: "{{ apache_config_path }}/{{ item }}"
+    mode: a+x
```
diurnalist, line 23: Thanks!

roles/prometheus/tasks/alertmanager.yml
```diff
@@ -8,6 +8,10 @@
     path: "{{ prometheus_alertmanager_config_dir }}"
     state: directory
 
+- name: Create Docker network.
```
diurnalist: I don't think you need this. Prometheus uses the default network and Alertmanager can also use that.

roles/prometheus/templates/alertmanager.service.j2
```diff
@@ -5,9 +5,12 @@ After=docker.service
 [Service]
 Restart=always
 ExecStart=/bin/docker run --name=%N \
+                          --net {{ prometheus_alertmanager_docker_network }} \
```
diurnalist: Can remove the `--net` part.

roles/prometheus/templates/alertmanager.service.j2
```diff
@@ -5,9 +5,12 @@ After=docker.service
...
+                          --publish {{ prometheus_alertmanager_port }}:9093 \
+                          --expose 9093 \
```
diurnalist: `--expose` is redundant when `--publish` is already being used. Can remove.

roles/prometheus/templates/alertmanager.service.j2
```diff
@@ -5,9 +5,12 @@ After=docker.service
...
+                          --mount 'type=bind,src=/var/run/docker.sock,dst=/var/run/docker.sock' \
```
diurnalist: Please remove this bind - we don't want to mount the Docker socket unless we absolutely have to (it is a security liability.) This was only relevant for JupyterHub, which launches Docker containers itself.

diurnalist (changes requested):

playbooks/prometheus.yml
jakecoll: Are there downsides to keeping?

diurnalist: No, it is save to keep.

diurnalist (approved):
Great stuff!

PR #7, opened 2018-11-19
Add gnocchi plugin to grafana docker

jakecoll:
This will add the gnocchi plugin to the grafana docker image.

### DIFF
--- a/roles/grafana/defaults/main.yml
+++ b/roles/grafana/defaults/main.yml
@@ -13,4 +13,5 @@ grafana_auth_ldap_group_treebase: ou=Chameleon,ou=Systems,dc=tacc,dc=utexas,dc=e
 grafana_auth_ldap_group_search_filter: (&(objectClass=groupOfNames)(member=uid=%s,ou=People,dc=tacc,dc=utexas,dc=edu))
 grafana_auth_ldap_editor_group: Chameleon
 
-grafana_gnocchi_plugin_id: gnocchixyz-gnocchi-datasource
+grafana_plugins:
+  - gnocchixyz-gnocchi-datasource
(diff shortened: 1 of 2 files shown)


roles/grafana/defaults/main.yml
```diff
@@ -12,3 +12,5 @@ grafana_auth_ldap_user_search_filter: (uid=%s)
 grafana_auth_ldap_group_treebase: ou=Chameleon,ou=Systems,dc=tacc,dc=utexas,dc=edu
 grafana_auth_ldap_group_search_filter: (&(objectClass=groupOfNames)(member=uid=%s,ou=People,dc=tacc,dc=utexas,dc=edu))
 grafana_auth_ldap_editor_group: Chameleon
+
+grafana_gnocchi_plugin_id: gnocchixyz-gnocchi-datasource
```
diurnalist: The "ansible way" of doing this is to use a list, given that what you ultimately have control over in the template is a _list_ of plugins to install.

```yaml
grafana_plugins:
  - gnocchixyz-gnocchi-datasource
```

Then in the template you transform that into the value you need... in this case it looks like GF_INSTALL_PLUGINS takes a [comma-separated list][1].

```
-e "GF_INSTALL_PLUGINS={{ grafana_plugins | join(',') }}"
```

[1]: https://github.com/grafana/grafana/blob/master/docs/sources/installation/docker.md#installing-plugins-for-grafana

diurnalist (changes requested):
This is a bit brittle if we want to add more plugins down the road, see suggestion!

diurnalist (approved):
🥇

diurnalist:
@jakecoll this looks good, can you please merge?

PR #8, opened 2018-11-27
Updated openstack_exporter for deployment at TACC and for gnocchi gpu…

jakecoll:
Updates openstack_exporter for deployment at TACC. This also adds the gnocchi endpoint to the docker env file.


diurnalist (approved):
Looks overall good, and thanks for bringing in the TACC changes. @francoishalbach: please be sure to commit the TACC inventory back here when you are perform any edits.

I think the Gnocchi endpoint is not really _necessary_ to have explicitly like this, given that OpenStack supports service discovery and so it should be possible and more future proof to have the exporter "discover" the Gnocchi endpoint itself. But this is kewl for now.

PR #10, opened 2019-01-28
[glance] Add glance playbook

zhenz:


roles/glance/templates/glance.conf.j2
```diff
@@ -0,0 +1,6 @@
+[keystone_authtoken]
+region_name = {{ openstack_region_name }}
```
diurnalist: Surprised that this region stuff is necessary - did the containers error out on start without this?

diurnalist (approved):
Looks good!

PR #11, opened 2019-01-29
Heat containerization

jakecoll:


kolla/node_custom_config/heat/policy.json
```diff
@@ -0,0 +1,100 @@
+{
```
diurnalist: I don't think we actually change the default policy. If that's true, we don't need this.

diurnalist (approved):

PR #12, opened 2019-02-04
update jenkins deployment of experiment precis

zhenz:


diurnalist (approved):
Nice stuff! I think the `network_interface_*` stuff shouldn't be necessary when I update the script to pull in the `globals.yml` file when it runs...

zhenz:
> Nice stuff! I think the `network_interface_*` stuff shouldn't be necessary when I update the script to pull in the `globals.yml` file when it runs...

I didn't see `network_interface_external` and `network_interface_internal` in `globals.yml`. But there are `neutron_external_interface` and `network_interface` in `globals.yml`. Should I update tasks yml files with the new variable names? If they are the same things.

diurnalist:
Oh nevermind! I was thinking about something else entirely. Forgot about this functionality in the ansible playbooks that is used to create the Apache proxies.

PR #13, opened 2019-02-15
Mysql replication alert

jakecoll:
Adds prometheus alert rules to notify alertmanager when a mysql slave replication error occurs.


roles/prometheus/defaults/main.yml
```diff
@@ -15,4 +15,4 @@ prometheus_alertmanager_users:
   - jason_a
 
 # This should be encrypted! It is a secret value.
-prometheus_alertmanager_slack_api_url: ''
+prometheus_alertmanager_slack_api_url: "{{ lookup('file', '/root/scripts/slack.json') | from_json }}"
```
diurnalist: This is somewhere in the `group_vars/` folder; you shouldn't have to set it here I think. Or, if it's not there, it should be as opposed to residing on a folder.

diurnalist: The reason is that it makes it harder to port this to a different server or something. I prefer that we keep only one server specific file: the vault password, which can be backed up in a myriad of ways and is only 1 file as opposed to N.

roles/prometheus/templates/alert-rules.yml.j2
```diff
@@ -0,0 +1,10 @@
+groups:
+- name: mysql
+  rules:
+  - alert: MySQLReplicationErro
```
diurnalist:
```suggestion
  - alert: MySQLReplicationError
```

roles/prometheus/templates/alert-rules.yml.j2
```diff
@@ -0,0 +1,10 @@
...
+    expr: mysql_slave_status_last_errno{hostname="fx2-01"} > 0
+    for: 30s
```
diurnalist: I would actually increase this to `5m` or something. 30 seconds could be a legit network issue/blip and a false positive.

roles/prometheus/templates/alert-rules.yml.j2
```diff
@@ -0,0 +1,10 @@
+groups:
+- name: mysql
+  rules:
+  - alert: MySQLReplicationErro
+    expr: mysql_slave_status_last_errno{hostname="fx2-01"} > 0
```
diurnalist: What labels are on this metric besides `hostname`? A nice thing Prometheus does is report labels on the metric for you with the alert. This means you don't need to actually do the `{hostname="fx2-01"}` thing. A better pattern is to alert on the entire metric, so that if the last_errno is nonzero for _any_ hostname, it fires. This makes it more general and we can re-use it at TACC for example.

roles/prometheus/templates/prometheus.yml.j2
```diff
@@ -12,7 +12,7 @@ alerting:
 
 # Load rules once and periodically evaluate them according to the global 'evaluation_interval'.
 rule_files:
-  - "/etc/prometheus/*-rules.yml"
+  - "{{ prometheus_config_dir }}/alert-rules.yml"
```
diurnalist: Maybe keep the `*-rules` in here and rename `alert-rules.yml` to `mysql-rules.yml`. Probably over time we will have multiple types of alerts for different things and one big file can get unwieldy.

diurnalist (changes requested):
A few suggestions, the main change we need is to the alert itself. Good stuff though 👍

diurnalist (approved):

PR #14, opened 2019-02-22
vendordata service

zhenz:


zhenz:
@diurnalist I didn't add/modify Jenkins. I think it would be better a part of nova deployment. The service is in https://github.com/ChameleonCloud/chameleon-vendordata.

roles/precis/defaults/main.yml
```diff
@@ -29,6 +29,7 @@ precis_mysql_external_database_privileges:
   - neutron.*:SELECT
   - nova.*:SELECT
   - nova_api.*:SELECT
+  - gnocchi.*:SELECT
```
diurnalist, line 32: ?

roles/vendordata/templates/chameleon_vendordata.service.j2
```diff
@@ -0,0 +1,18 @@
+# {{ ansible_managed }}
+[Unit]
+Description=Chameleon vendordata service
+After=docker.service
+
+[Service]
+Restart=always
+ExecStart=/bin/docker run --name %N \
+                          --net {{ vendordata_docker_network }} \
```
diurnalist: I think we are going to have to have `--net=host` here instead. The openstack services currently on the host, and even with Kolla, are binding on the host's network interfaces. A docker network will effectively hide the vendordata service from the rest of the system.

roles/vendordata/templates/chameleon_vendordata.service.j2
```diff
@@ -0,0 +1,18 @@
...
+                          --publish '{{ vendordata_port }}:8080' \
```
diurnalist: You don't need this if you're using `--net=host` I believe.

roles/vendordata/templates/vendordata_dynamic.conf.j2
```diff
@@ -0,0 +1,12 @@
+[openstack]
+auth_url: {{ openstack_keystone_auth_url }}
```
diurnalist: Is this a valid conf format? I thought it was usually like `key = value`, not `key: value`

roles/vendordata/templates/vendordata_static.json.j2
```diff
@@ -0,0 +1,15 @@
+{
+	"auth_url": "{{ openstack_keystone_auth_url }}",
+	"auth_url_v2": "{{ openstack_keystone_auth_url }}/v2.0",
+	"auth_url_v3": "{{ openstack_keystone_auth_url }}/v3",
+	"cluster": "chameleon",
+	"grid": "chameleoncloud",
+	"site": "{{ chameleon_site_name }}",
+	"region": "{{ openstack_region_name }}",
+	"rabbit_host": "{{ rabbit_host }}",
```
diurnalist: I think you can drop all the rabbit stuff. I couldn't find anywhere this was used. I think it was an initial implementation of the metrics collection.

kolla/node_custom_config/nova.conf
```diff
@@ -45,3 +45,17 @@ max_attempts = 50
 
 [serial_console]
 base_url = wss://{{ kolla_external_fqdn }}:{{ nova_serialproxy_port }}/
+
+[api]
+vendordata_jsonfile_path = "{{ vendordata_config_path }}/vendordata_static.json"
+vendordata_providers = StaticJSON,DynamicJSON
+vendordata_dynamic_targets = instance_metrics_writer_token@http://127.0.0.1:8911/instance_metrics_writer_token,node@http://127.0.0.1:8911/node,project_id@http://127.0.0.1:8911/project_id,user_id@http://127.0.0.1:8911/user_id
```
diurnalist: My understanding from reading the docs is that you only define this once per target. So it would be like this instead:

```
vendordata_dynamic_targets = chameleon@http://127.0.0.1:8911
```

The vendordata2 endpoint would then return something like:
```json
{
  "chameleon": {
    "node": "...",
    "user_id": "...",
    "instance_metrics_writer_token": "...",
  }
}
```

diurnalist (changes requested):
This is looking very fleshed out, almost complete! Nice work. Left some comments about the configuration of the network and the Nova conf.

roles/precis/defaults/main.yml
zhenz: Oops! I kept merging master, and it must be something I messed up with. I'll delete.

roles/vendordata/templates/vendordata_dynamic.conf.j2
zhenz: It can be read by ConfigParser. I have this format for precis.

kolla/node_custom_config/nova.conf
zhenz: I had a second thought right after I created this pull request. I agree with you. Making changes now.

roles/precis/defaults/main.yml
zhenz: Ah! I remembered that I added that. But this is irrelevant to vendordata. We need gnocchi database select previledge for precis serice.

zhenz:
@diurnalist I'm done with the fixes. I have it tested on dev@tacc.

diurnalist (approved):
Awesome! 

nicejob

PR #18, opened 2019-03-13
Support for corsa stats in prometheus openstack exporter

jakecoll:
Primarily just adds a config file for openstack_exporter to access corsa info in prometheus role.


roles/prometheus/defaults/main.yml
```diff
@@ -80,3 +80,5 @@ prometheus_services:
     config_dir: /etc/prometheus
     env_file: /etc/prometheus/openstack_exporter.env
     scrape_target: yes
+    volumes:
+      - "/etc/prometheus/openstack_exporter.yml:/usr/local/bin/exporte/config.yml"
```
diurnalist: `exporte` <= typo?

roles/prometheus/tasks/openstack_exporter.yml
```diff
@@ -1,6 +1,16 @@
...
+
+- name: Set facts from global configuration.
+  set_fact:
+    openstack_exporter_corsa_switches: "{{ switch_configs | selectattr('device_type', 'match', '^corsa') | list }}"
+
+- name: Configure Openstack exporter config file.
+  template:
+    src: openstack_exporter.yml.j2
+    mode: 0600
+    dest: "{{ service.config_dir}}/openstack_exporter.yml"
```
diurnalist:
```suggestion
    dest: "{{ service.config_dir }}/openstack_exporter.yml"
```

roles/prometheus/defaults/main.yml
diurnalist: Needs to be `/root/openstack_exporter.yml` actually, given how you define it in `OS_EXPORTER_CONF`.

roles/prometheus/templates/openstack_exporter.yml.j2
```diff
@@ -0,0 +1,14 @@
+---
+{% if openstack_exporter_corsa_switches | length > 0 %}
+switch_corsa:
+  resource_type: switch
+  # Batch size is roughly the number of ports across all switches
+  batch_size: {{ (openstack_exporter_corsa_switches | length) * 32 }}
```
diurnalist: `resource_type` and `batch_size` have no meaning for your exporter from what I can tell.

diurnalist (approved):
Some stuff that can be cleaned up, overall makes sense 👍

PR #20, opened 2019-03-26
Add snmp exporter

jakecoll:


roles/prometheus/templates/snmp_generator.yml.j2
```diff
@@ -0,0 +1,12 @@
+---
+{% if snmp_exporter_switches | length > 0 %}
```
diurnalist: Instead of this length check here, I would check it in the task itself and do something like

```
when: snmp_exporter_switches | length > 0
```

You could add that block to both the generator config file and the run of the generator itself.
Another thing you could do is trigger the 'generate config file for snmp exporter' part only when the template changed. That is possible via something like:

```
name: Configure snmp exporter' generators config file
template: ...
register: snmp_exporter_generator_config

name: Generate config file
docker_container: ...
when: snmp_exporter_generator_config.changed
```

Not 100% if that works, but could be nicer?

diurnalist (approved):
Cool! I had to look at what the MIBs stuff was doing. Looks like it's just pulling down specs about how metrics are scraped from various switches? Seems legit.

PR #23, opened 2019-04-03
Remove vlan ranges for 1 exogeni and 60 physnet1 at UC and TACC

jakecoll:
Verified neither of the network segments removed are in use. Should be able to reclaim one exogeni segment at TACC and add the range of 60 vlan tags to blazar for each site.


diurnalist (approved):
Sounds good, we should remove those old exogeni nets but 👍

PR #24, opened 2019-04-09
Add rule to fire alert for instance launch failures.

jakecoll:


diurnalist (approved):
This is likely to be very chatty, but we can try it out. I imagine the threshold will have to be updated to be something like, 5? idk.

PR #25, opened 2019-04-11
Use external site configs

diurnalist:
This updates our Ansible tooling to use external site configurations (which are now in the ChameleonCloud/site-configs repository).

This also simplifies the tooling such that everything can be driven by one `cc-ansible` script, deprecating the old 2-script solution.


codyhammock (approved):
Looks good!

PR #27, opened 2019-04-24
Hammers

jakecoll:
Adds playbook and roles for Hammers

### DIFF
--- a/roles/hammers/tasks/main.yml
+++ b/roles/hammers/tasks/main.yml
@@ -1,0 +1,52 @@
+---
+- set_fact:
+    hammers_list: "{{ hammers | dict2items | map(attribute='value') | list }}"
+
+- name: Pull Docker image.
+  docker_image:
+    name: "{{ hammers_docker_image }}"
+
+- name: Create config directory.
+  file:
+    path: "{{ hammers_config_path }}"
+    state: directory
+
+- name: Configure Hammers docker bash wrapper
+  template:
+    src: cc-hammer.j2
+    mode: a+x
+    dest: /usr/bin/cc-hammer
+
+- name: Configure slack webhook file
+  template:
+    src: slack.json.j2
+    mode: 0600
+    dest: "{{ hammers_config_path }}/slack.json"
+
+- name: Create MySQL user
+  local_action:
+    module: mysql_user
+    login_host: "{{ groups.mariadb[0] }}"
+    login_password: "{{ database_password }}"
+    name: "{{ hammers_mysql_user }}"
+    host: 172.%
+    password: "{{ hammers_mysql_password }}"
+    priv: "*.*:SELECT,UPDATE,DELETE"
+
+- name: Configure Hammers my.cnf
+  template:
+    src: my.cnf.j2
+    dest: "{{ hammers_config_path }}/my.cnf"
+  vars:
+    mysql_host: "{{ ansible_internal_ipv4_addresses[groups.mariadb[0]] }}"
+
+- name: Add hammer cron jobs.
+  cron:
+    name: "{{ item.hammer_name }}"
+    minute: "{{ item.cron_times.minute | default('*') }}"
+    hour: "{{ item.cron_times.hour | default('*') }}"
+    day: "{{ item.cron_times.day | default('*') }}"
+    month: "{{ item.cron_times.month | default('*') }}"
+    weekday: "{{ item.cron_times.weekday | default('*') }}"
+    job: "{{ item.cmd }} {{ item.extra_args | default([]) | join(' ') }}"
+  loop: "{{ hammers_list }}"
--- a/roles/hammers/tasks/main.yml
+++ b/roles/hammers/tasks/main.yml
@@ -5,6 +5,7 @@
 - name: Pull Docker image.
   docker_image:
     name: "{{ hammers_docker_image }}"
+    force: yes
 
 - name: Create config directory.
   file:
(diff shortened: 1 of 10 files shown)


roles/hammers/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
+---
+- set_fact:
+    hammers_list: "{{ hammers | dict2items | map(attribute='value') | list }}"
+
+- name: Pull Docker image.
+  docker_image:
+    name: "{{ hammers_docker_image }}"
```
diurnalist, line 7: Might want to add `force: yes` here to ensure that it re-pulls even if the tag hasn't changed (which it won't if we're always just using `latest` as the tag.)

roles/hammers/templates/hammers.env.j2
```diff
@@ -0,0 +1,11 @@
+OS_AUTH_STRATEGY=keystone
+OS_AUTH_URL={{ openstack_keystone_auth_url }}
+OS_USERNAME={{ prometheus_openstack_exporter_user }}
+OS_PASSWORD={{ prometheus_openstack_exporter_password }}
+OS_PROJECT_NAME=openstack
+OS_PROJECT_ID={{  }}
```
diurnalist: typo?

roles/hammers/templates/hammers.env.j2
```diff
@@ -0,0 +1,11 @@
+OS_AUTH_STRATEGY=keystone
+OS_AUTH_URL={{ openstack_keystone_auth_url }}
+OS_USERNAME={{ prometheus_openstack_exporter_user }}
```
diurnalist: Please do not use variables set outside of your role - instead, do like `{{ hammers_keystone_user }}` and `{{ hammers_keystone_password }}`. Better to keep all the stuff that can affect how a role works internal to the role itself. It will not be obvious to somebody else that changing the prometheus exporter password will require also re-running the hammers provisioning step (so that the password is updated for the containers.)

roles/hammers/templates/slack.json.j2
```diff
@@ -0,0 +1,8 @@
+{
+    "webhook": "{{ slack_hammers_webhook }}",
+    "hostname_names": {
```
diurnalist, line 3: What is this for?

roles/hammers/templates/cc-hammer.j2
```diff
@@ -0,0 +1,13 @@
+#! /bin/sh
+
+slack="{{ hammers_config_path }}/slack.json"
+log="/usr/bin/logger"
+log_tag="hammers-"$1
+
+docker run -it \
+  -e "{{ hammers_env_file }}" \
+  -v "/root/adminrc:/etc/hammers/adminrc" \
```
diurnalist: Do you need this if you're already passing in the `hammers_env_file` as environment variables? I'd think only one would be needed (and I think passing in env is the way to go!)

roles/hammers/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+
+- name: Add hammer cron jobs.
+  cron:
+    name: "{{ item.hammer_name }}"
+    minute: "{{ item.cron_times.minute | default('*') }}"
+    hour: "{{ item.cron_times.hour | default('*') }}"
+    day: "{{ item.cron_times.day | default('*') }}"
+    month: "{{ item.cron_times.month | default('*') }}"
+    weekday: "{{ item.cron_times.weekday | default('*') }}"
+    job: "{{ item.cmd }} {{ item.extra_args | default([]) | join(' ') }}"
```
diurnalist: It seems simpler to just include the entire command on one line, rather than trying to assemble its args in this way. If you don't need to process `extra_args` in any other way, I'd suggest doing that.

roles/hammers/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+- name: Create config directory.
+  file:
+    path: "{{ hammers_config_path }}"
+    state: directory
+
+- name: Configure Hammers docker bash wrapper
+  template:
+    src: cc-hammer.j2
+    mode: a+x
+    dest: /usr/bin/cc-hammer
```
diurnalist: I'd put this in `/usr/local/sbin`. `/usr/local` is a place for system-wide scripts that are not included as part of the distro (yum). `sbin` is for things that have to run as root -- this script is a candidate because it needs Docker, which requires root.

For a quick read on the difference between all these path conventions, check out [this answer](https://unix.stackexchange.com/a/8658).

diurnalist (changes requested):
Can you update this so that the hammers user/password are somehow set in the role?

Also, have a look at the Git history; it looks like a messy merge may have occurred. You could try a `git pull --rebase` on your branch to fix.

roles/hammers/templates/slack.json.j2
jakecoll: this is how hammers sends slack notifications.

roles/hammers/templates/cc-hammer.j2
```diff
@@ -0,0 +1,13 @@
+#! /bin/sh
+
+slack="{{ hammers_config_path }}/slack.json"
+log="/usr/bin/logger"
+log_tag="hammers-"$1
+
+docker run -it \
```
diurnalist: FYI I don't think you need `-i` or `-t` here. `-i` is mostly used when you need to use STDIN (pipe something in to the program or have an interactive input by typing stuff) and `-t` allocates a TTY, which is needed to allow typing as if it's a "regular" console.

roles/hammers/templates/cc-hammer.j2
```diff
@@ -0,0 +1,13 @@
...
+  -e "{{ hammers_env_file }}" \
```
diurnalist: You have to say `--env-file` if you're passing in a file. `-e` (short for `--env`) is for when you're passing in a single variable like `-e 'FOO=bar'`

diurnalist (approved):
Fix the `--env-file` (it won't work without this, I wouldn't think), but then :shipit: !

PR #28, opened 2019-06-04
[chameleon_usage] add playbook for chameleon usage

zhenz:


roles/chameleon_usage/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
...
+- name: Pull Docker image.
+  docker_image:
+    name: "{{ chameleon_usage_docker_image }}"
+    force: yes
+    
+- name: Initialize chameleon_usage database
+  shell: |
+    docker run --rm --net=host \
+    -v "{{ chameleon_usage_config_dir }}/.my.cnf:/root/.my.cnf" \
+    -v "/var/lib/mysql/mysql.sock:/var/lib/mysql/mysql.sock" \
```
diurnalist: I don't think you should have to mount the socket in here. Can you access it via its private IP instead? That's how all the OpenStack services do it.

diurnalist (commented):
Is there a dedicated MySQL user for this database?

zhenz:
> Is there a dedicated MySQL user for this database?

It creates database/tables, dumps backups, queries, and writes records as root. Maybe need a grafana user with read-only access to chameleon-usage database? Do we have one already?

roles/chameleon_usage/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+    append_privs: 'yes'
+  with_items:
+    - 'blazar_backup\_%.*:CREATE,DROP,SELECT,INSERT'
+    - 'keystone_backup\_%.*:CREATE,DROP,SELECT,INSERT'
+    - 'nova_backup\_%.*:CREATE,DROP,SELECT,INSERT'
+    - 'chameleon_usage.*:CREATE,SELECT,INSERT,UPDATE'
+  vars:
+    # Ensure when we run this on the ansible host, we use the Virtualenv
+    # (which should already be set up.)
+    ansible_python_interpreter: "{{ ansible_python_interpreter }}"
```
diurnalist: I don't think this does anything. The variable should already be in scope (otherwise you wouldn't be able to set it!)

roles/chameleon_usage/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+    src: my.cnf.j2
+    dest: "{{ chameleon_usage_config_dir }}/my.cnf"
+  vars:
+    mysql_host: "{{ chameleon_usage_mysql_host }}"
+    root_password: "{{ database_password }}"
+    
+- name: Initialize chameleon_usage database
+  shell: |
+    docker run --rm --net=host \
+    -v "{{ chameleon_usage_config_dir }}/my.cnf:/root/.my.cnf" \
```
diurnalist: Nit picking, but I think [`/etc/mysql/my.cnf`](https://mariadb.com/kb/en/library/configuring-mariadb-with-option-files/#default-option-file-locations-on-linux-unix-mac) might be the more appropriate place for this. Otherwise if we eventually change the executing user (for some reason), it would break. The `/etc/mysql` folder should set this globally, for any user.

roles/chameleon_usage/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+    {{ chameleon_usage_docker_image }} setup_database 2>&1
+    
+- name: Configure chameleon usage data extract docker bash wrapper
+  template:
+    src: chameleon-usage-extract-data.j2
+    mode: a+x
+    dest: /usr/local/sbin/chameleon-usage-extract-data
+    
+- name: Fill in up-to-date data to chameleon_usage
+  shell: "/usr/local/sbin/chameleon-usage-extract-data"
```
diurnalist: Use `command` for this, it is a bit safer. `shell` is only really for when you need the shell to do variable expansion or other "shell"-y things.

roles/chameleon_usage/templates/chameleon-usage-extract-data.j2
```diff
@@ -0,0 +1,7 @@
+#! /bin/sh
+
+docker run --rm --net=host \
+    -v "{{ backup_dir }}:{{ backup_dir }}" \
+    -v "{{ chameleon_usage_config_dir }}/my.cnf:/root/.my.cnf" \
```
diurnalist: Same comment as above re: the my.cnf location

diurnalist (approved):
This looks great, very clean! I left some comments about small things you could do to improve it, but they are minor.

awesome

PR #30, opened 2019-07-02
Configure mariadb, glance, and gnocchi backups

hlbfrancois:


roles/glance/templates/glance.j2
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+cd "{{ glance_backup_origin_dir }}" && \
```
diurnalist, line 3: I'm not 100% sure if it's the right thing, but one way to reduce the amount of extra configuration you have to set here is having defaults for most/all of this.

```yaml
glance_backup_origin_dir: "{{ glance_file_datadir_volume }}"
glance_backup_swift_endpoint: "{{ swift_public_endpoint }}"
glance_backup_container_name: glance_backup
```

Another thing... you could generate the glance token on every run. Probably a good idea to rotate this frequently. Also, it helps because these tokens _can_ expire. And in fact, they usually do with much greater regularity than we have currently configured. You can do this using the [`os_auth`](https://docs.ansible.com/ansible/latest/modules/os_auth_module.html) module, from a bit of research.

```yaml
- name: Generate auth token for Glance backup
  os_auth:
    auth:
      username: "{{ glance_backup_username }}"
      password: "{{ glance_backup_keystone_password }}"
      project_name: "{{ glance_backup_project }}"
      domain_name: "default"
      user_domain_name: "default"
  register: glance_backup_auth

- name: Set fact for Glance auth token
  set_fact:
    glance_backup_auth_token: "{{ glance_backup_auth.auth_token }}"
```

roles/gnocchi/templates/gnocchi.j2
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+export SWIFT_PREAUTHTOKEN="{{ gnocchi_backup_auth_token }}" && \
```
diurnalist, line 3: Same comments about token/auth/defaults apply for this one.

diurnalist (approved):
I would get rid of all the boilerplate files, like the README, the meta files, the tests... anything that is not relevant to this repo/code working adds to confusion later on, in my opinion.

I also think that looking at generating auth tokens on-demand is a good idea for long-term maintainability.

roles/glance/meta/main.yml
```diff
@@ -0,0 +1,60 @@
+galaxy_info:
```
diurnalist, line 1: Please remove this file

roles/glance/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
+---
+- name: Generate auth token for Glance backup
+  os_auth:
+    auth:
+      username: "{{ glance_backup_username }}"
```
diurnalist: Nice 👍 - so you made a new user for this?

roles/gnocchi/meta/main.yml
```diff
@@ -0,0 +1,60 @@
+galaxy_info:
```
diurnalist, line 1: Please remove

roles/mariadb/meta/main.yml
```diff
@@ -0,0 +1,60 @@
+galaxy_info:
```
diurnalist, line 1: Please remove

diurnalist (approved):
Very nice!

roles/glance/tasks/main.yml
hlbfrancois: yes ! and deleted the old one :) thanks

PR #32, opened 2019-07-23
[ironic] increase neutron timeout from default 30 seconds to 900 seconds

zhenz:
When launching a large amount of nodes, typically some will fail provisioning. The error is due to the timeout when trying to create the network port.


diurnalist (approved):
I think this PR should probably have the 15m server timeout also defined; they go together (should be the same value I think)

diurnalist:
Thanks @zhenz ! perfect

zhenz:
> I think this PR should probably have the 15m server timeout also defined; they go together (should be the same value I think)

Do you want me to put a comment in ironic.conf file or update the commit message? Sorry, I'm not quite sure I understand your comment.

PR #36, opened 2019-10-02
Adding noop driver to allow for static boot order configurations

barnesalex:
tested this in dev with no issues

### DIFF
--- a/kolla/node_custom_config/ironic.conf
+++ b/kolla/node_custom_config/ironic.conf
@@ -6,6 +6,9 @@ default_network_interface = neutron
 my_ip = "{{ kolla_internal_vip_address }}"
 enabled_storage_interfaces = cinder,noop
 default_storage_interface = cinder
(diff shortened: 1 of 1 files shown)


codyhammock (approved):
Appears to be what's in DEV@TACC, and is consistent with documentation.

Looks good! 👍

kolla/node_custom_config/ironic.conf
```diff
@@ -6,6 +6,9 @@ default_network_interface = neutron
 my_ip = "{{ kolla_internal_vip_address }}"
 enabled_storage_interfaces = cinder,noop
 default_storage_interface = cinder
+enabled_management_interfaces = ipmitool,noop
+default_management_interface = ipmitool
```
diurnalist, line 10: FYI the first enabled iface will be the default, according to the docs. Good to be explicit though.

diurnalist (approved):
LGTM

PR #37, opened 2019-10-04
[chameleon usage] add heat template download reporting

zhenz:


roles/chameleon_usage/tasks/main.yml
```diff
@@ -58,3 +58,25 @@
     month: "*"
     weekday: "*"
     job: "/usr/local/sbin/chameleon-usage-extract-data | /usr/bin/logger -t \"chameleon_usage\""
+    
+- name: Configure chameleon usage heat template download docker bash wapper
+  when: openstack_region_name == 'CHI@UC'
```
diurnalist: If this is something conditionally enabled on sites, I think it's better to have it behind a flag like `enable_usage_appliance_report: no` - that can be in kolla/defaults.yml. the UC site config can then set this to `yes`. This makes it easier for operators to toggle this on/off; it's not implicit based on the region name that way.

diurnalist (approved):
Looks good to me! I think in the future we might want to consolidate these non-site-specific reporters to one site. Just need to pick one.

PR #38, opened 2019-10-09
[neutron] Add jumbo frame configs

jakecoll:
These configs will set a default of 9000 for any neutron network created on top of exogeni or physnet.

### DIFF
--- a/kolla/node_custom_config/neutron/ml2_conf.ini
+++ b/kolla/node_custom_config/neutron/ml2_conf.ini
@@ -2,6 +2,13 @@
 tunnel_types =
 
 [ml2]
+{% if neutron_networks is defined %}
+physical_network_mtus = {% for config in neutron_networks %}{{ config.name }}:{{ config.mtu }}{% if not loop.last %},{% endif %}{% endfor %}
+{% elif neutron_ovs_bridge_mappings is defined %}
+physical_network_mtus = {{ neutron_physical_network_mtus }}
+{% endif %}
+
+path_mtu = 9000
 mechanism_drivers = openvswitch,genericswitch{% if enable_ironic_neutron_agent | bool %},baremetal{% endif %}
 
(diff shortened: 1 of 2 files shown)


kolla/node_custom_config/neutron/ml2_conf.ini
```diff
@@ -2,6 +2,13 @@
 tunnel_types =
 
 [ml2]
+{% if neutron_networks is defined %}
+physical_network_mtus = {% for config in neutron_networks %}{{ config.name }}:{{ config.mtu }}{% if not loop.last %},{% endif %}{% endfor %}
+{% elif neutron_ovs_bridge_mappings is defined %}
```
diurnalist, line 7: I don't know that you need to have this `elif` - newer config items should be using `neutron_networks` as you have it here. Nice making this overridable on the per-provider-net level though!

kolla/node_custom_config/neutron/ml2_conf.ini
```diff
@@ -2,6 +2,13 @@
...
+physical_network_mtus = {{ neutron_physical_network_mtus }}
+{% endif %}
+
+path_mtu = 9000
```
diurnalist, line 11: After reading what this does, I don't think it applies to us: https://docs.openstack.org/neutron/rocky/configuration/ml2-conf.html#ml2.path_mtu

diurnalist (commented):
Not sure what to do about this given some feedback from @paul-ruth about being careful where fragmentation might be occurring. If it happens in OVS on the Neutron node, that could make Neutron even more overburdened. We could leave the settings the way they currently are and simply update documentation to instruct users how to do `openstack network set --mtu 9000 <network>` for their own networks.

diurnalist:
We decided not to do this in lieu of documenting how users can set this themselves.

PR #41, opened 2019-11-05
[horizon] update configuration settings

zhenz:


zhenz:
@diurnalist @xandroc OPENSTACK_HYPERVISOR_FEATURES doesn't work based on some bug reports (I tested as well), such as https://lists.launchpad.net/yahoo-eng-team/msg72417.html. I still include it anyways. Since we have serial console enabled, key pair might be ok to be not required.

kolla/node_custom_config/horizon/custom_local_settings
```diff
@@ -70,3 +70,36 @@ CHAMELEON_PORTAL_SSO_LOGIN_PATH = '/sso/horizon/'
 CHAMELEON_PORTAL_SSO_LOGOUT_PATH = '/logout/'
 SSO_CALLBACK_HOST = '{{ kolla_external_fqdn }}'
 {% endif %}
+
+# A dictionary of settings which can be used to provide the default values for
+# properties found in the Launch Instance modal.
+LAUNCH_INSTANCE_DEFAULTS = {
```
diurnalist, line 76: Very cool! Is it possible to specify a default flavor, do you know?

kolla/node_custom_config/horizon/custom_local_settings
```diff
@@ -70,3 +70,36 @@ CHAMELEON_PORTAL_SSO_LOGIN_PATH = '/sso/horizon/'
...
+        ('qcow2', 'QCOW2 - QEMU Emulator'),
+        ('raw', 'Raw'),
+    ],
+}
+
+# The Xen Hypervisor has the ability to set the mount point for volumes
+# attached to instances (other Hypervisors currently do not). Setting
+# can_set_mount_point to True will add the option to set the mount point
+# from the UI.
+OPENSTACK_HYPERVISOR_FEATURES = {
```
diurnalist: How does this apply to baremetal? And, this might affect the KVM deployment, which also uses CHI-in-a-box @hammock

diurnalist (approved):
Looks good from my perspective, let's see what others think.

kolla/node_custom_config/horizon/custom_local_settings
zhenz: The only param about flavor is CREATE_INSTANCE_FLAVOR_SORT https://docs.openstack.org/horizon/rocky/configuration/settings.html#create-instance-flavor-sort, which is used to sort the flavors when launching an instance. I tried to set the key to name and hope the first one (we only have one) would be the default selection, but doesn't work.

kolla/node_custom_config/horizon/custom_local_settings
zhenz: I tested OPENSTACK_HYPERVISOR_FEATURES and it doesn't work. I can take it out.

diurnalist: I would take it out; it will probably mess up the KVM site.

codyhammock (approved):
Looks pretty straightforward 👍

kolla/node_custom_config/horizon/custom_local_settings
codyhammock: KVM overrides this in its own branch. I wouldn't worry about it.

xandroc:
Looks good @zhenz nice work!

xandroc (approved):

PR #44, opened 2019-11-13
[prometheus] Use built-in Ceph exporters

diurnalist:
Ceph has the ability to enable Ceph exporters natively[1]. This
leverages that functionality as opposed to installing new exporters on
the node ourselves.

It is already possible to define jobs that don't provision/install a
service, but rather assume that a port is already exposing metrics.

- This will remove the 'cluster' label, but it doesn't look like we
really need that.

[1]: https://docs.ceph.com/docs/master/mgr/prometheus/

### DIFF
--- a/roles/prometheus/defaults/main.yml
+++ b/roles/prometheus/defaults/main.yml
@@ -96,6 +96,14 @@ prometheus_services:
     scrape_target: no
     volumes:
      - "/etc/prometheus/snmp-exporter:/etc/snmp_exporter"
+  ceph-exporter:
+    group: prometheus-ceph-exporter
+    port: 9283
+    scrape_target: yes
+  ceph-node-exporter:
+    group: prometheus-ceph-exporter
+    port: 9100
+    scrape_target: yes
   push-gateway:
     service_name: prometheus_pushgateway
     image: prom/pushgateway:v0.8.0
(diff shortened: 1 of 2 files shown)


roles/prometheus/defaults/main.yml
```diff
@@ -96,6 +96,14 @@ prometheus_services:
     scrape_target: no
     volumes:
      - "/etc/prometheus/snmp-exporter:/etc/snmp_exporter"
+  ceph-exporter:
```
diurnalist, line 99: The prometheus role already has support for looping over all the services defined here, and in the event of there being no `image` or `service_name` defined, it will simply add entries to the prometheus configuration to scrape these hosts on this port.

codyhammock:
Our grafana dashboard does use the cluster label, e.g. `ceph_cluster_total_bytes{cluster="$region"}`. This label can be useful if we wish to have a combined dashboard to provide a view of multiple sites.

diurnalist:
@codyhammock, I don't quite understand. We already have a combined dashboard, where you can switch sites by specifying which datasource (which Prometheus server) is providing the data. Each site has its own Prometheus server by design.

codyhammock:
After experimentation in Grafana, it appears that the `cluster` label is unnecessary, after all.

PR #45, opened 2019-11-15
Cleanup cron notifications on backup scripts

diurnalist:
Instead of defining the same notification script in multiple places, all
of which will override others when they run on the same host, pull that
part out to a common task and include it from the other roles.

Also updates the notification script to include some helpful information
e.g. the type and help text for the metric, which Prometheus needs in
order to make sense of this.

### DIFF
--- a/roles/common/defaults/cron.yml
+++ b/roles/common/defaults/cron.yml
@@ -1,0 +1,2 @@
+---
+cron_notification: /usr/local/sbin/notification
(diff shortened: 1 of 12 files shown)


diurnalist:
I noticed some errors in the push gateway that were preventing scrapes from working; it seemed to want the 'TYPE' declaration so it knew this was a gauge metric.

hlbfrancois (approved):
Thanks! That makes sense.

PR #47, opened 2019-12-02
[chameleon_usage] jupyter usage

zhenz:


playbooks/chameleon_usage.yml
```diff
@@ -1,4 +1,26 @@
...
+      vars:
+          action: 'heat_template'
+
+- hosts: chameleon_usage_jupyter
+  roles:
+    - role: chameleon_usage
+      vars:
+          action: 'config'
+    - role: chameleon_usage
+      when: enable_usage_jupyter_report
```
diurnalist: Is it the case that there would be a host marked as `chameleon_usage_jupyter` but you wouldn't want `enable_usage_jupyter_report`? Perhaps we don't need this variable.

diurnalist (approved):
Nice splitting of the role into its component tasks. Feels like there is some boilerplate there that could be reduced, but it wasn't obvious to me how :D

👍

PR #48, opened 2019-12-05
Update to make stress test more flexible

barnesalex:
Utilizing native ansible fact info to make the test more flexible for different mem sized nodes.


diurnalist (approved):
Nice one

codyhammock (approved):
👍

PR #50, opened 2020-01-14
[common] Add restart fluentd cron job

jakecoll:
Centralized logging fails when ElasticSearch can't connect to fluentd. This
solution is a bit hacky, but restarting fluentd every few hours seems to
solve the issue.

Wasn't sure this was the best place to put this...

### DIFF
--- a/playbooks/cron.yml
+++ b/playbooks/cron.yml
@@ -14,3 +14,10 @@
         job: "/usr/bin/docker restart fluentd | /usr/bin/logger -t {{ job_name }} && {{ cron_notification }} {{ job_name }}"
       vars:
         job_name: restart-fluentd
+
+- hosts: elasticsearch
+  tasks:
+    - name: Add ElasticSearch cron job
+      include_role:
+        name: common
+        tasks_from: elasticsearch
--- a/playbooks/cron.yml
+++ b/playbooks/cron.yml
@@ -1,0 +1,16 @@
+---
+- hosts: baremetal
+  tasks:
+    - name: Add common cron jobs
+      include_role:
+        name: common
+        defaults_from: cron
+        tasks_from: cron
+    - name: Add fluentd restart job
+      cron:
+        name: "{{ job_name }}"
+        minute: 0
+        hour: "*/6"
+        job: "/usr/bin/docker restart fluentd | /usr/bin/logger -t {{ job_name }} && {{ cron_notification }} {{ job_name }}"
+      vars:
+        job_name: restart-fluentd
--- a/roles/common/tasks/elasticsearch.yml
+++ b/roles/common/tasks/elasticsearch.yml
@@ -1,0 +1,14 @@
+---
+- name: Add ElasticSeach clean up script
+  template:
+    src: clean_es_curator.sh.j2
+    dest: /usr/local/sbin/clean_es_curator
+    mode: 0700
+  vars:
+    es_curator_retention_days: 5
+- name: Add ElasticSearch clean up job
+  cron:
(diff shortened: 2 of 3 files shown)


roles/common/tasks/cron.yml
```diff
@@ -7,8 +7,28 @@
   vars:
     prometheus_push_gateway: "http://{{ groups['prometheus-push-gateway'][0] }}:9091"
 
+- name: Add ElasticSeach clean up script
```
diurnalist: The cron tasks are intended to be a general set of tasks that support having a cronjob. Placing this here will mean that this cronjob gets installed on every node, which isn't what we want.

roles/common/defaults/cron.yml
```diff
@@ -1,2 +1,18 @@
 ---
 cron_notification: /usr/local/sbin/notification
+cron_es_curator: /usr/local/sbin/clean_es_curator
+es_curator_retention_days: 5
+
+cron_jobs:
+  restart_fluentd:
+    job_name: "restart-fluentd"
+    cmd: "docker restart fluentd"
```
diurnalist: In general cronjobs should have the binary prefixed with the full path, e.g. `/usr/bin/docker` here.

playbooks/cron.yml
```diff
@@ -0,0 +1,8 @@
+---
+- hosts: common
```
diurnalist: `common` is not defined in the Ansible inventory anywhere (the `hosts` file), so I think this won't work -- did you test it?

I think you can probably target this better by doing the following, keeping it all within a playbook:

```yaml
---
- hosts: baremetal
  tasks:
    - name: Add common cron jobs
      include_role:
        name: common
        defaults_from: cron
        tasks_from: cron
    - name: Add fluentd restart job
      cron:
        minute: 0
        hour: "*/6"
        job: "docker restart fluentd | /usr/bin/logger -t {{ job_name }} && {{ cron_notification }} {{ job_name }}"
      vars:
        job_name: restart-fluentd
```

That's probably the cleanest place to keep it for now. It's a bit awkward b/c you have to know that the cron.yml playbook affects something like fluentd. Better for this to really live in kolla-ansible, but let's get something out to start.

playbooks/cron.yml
```diff
@@ -0,0 +1,8 @@
...
+  tasks:
+    - name: Add common cron jobs
+      include_role:
+        name: common
+        defaults_from: cron
+        tasks_from: cron
```
diurnalist, line 8: Following the previous pattern the ES stuff would then be

```yaml
- hosts: elasticsearch
  tasks:
    # 1. Template out the ES curator script
    # 2. Add the cron job
```

diurnalist (changes requested):
I don't think we should be bundling so much logic on the main cron tasks, which are run for effectively every host. See suggestions on how we might tackle this at the playbook level.

playbooks/cron.yml
jakecoll: I believe I added a common to site-configs on dev.

playbooks/cron.yml
jakecoll: The problem I run into with this one is that ansible can't seem to find the path of the curator's jinja script. The easier fix was to create an elasticsearch tasks file and include the role. This isn't great for consistency, but it does seem to do the trick.

playbooks/cron.yml
jakecoll: This can be done, but the ES curator doesn't play nicely with this structure because of the template.

diurnalist (approved):
nailedit

PR #51, opened 2020-01-15
Usage reporting

hlbfrancois:


diurnalist (changes requested):
For simplicity, let's roll all of this into the existing `chameleon_usage` roll. It already has a playbook defined that you can use. Perhaps ask @zhenz how it can work to add new items.

playbooks/chameleon_usage.yml
```diff
@@ -14,6 +14,10 @@
       when: enable_usage_appliance_report
       vars:
           action: 'heat_template'
+    - role: chameleon_usage
+      when: enable_usage_report_sync
```
diurnalist: We need to add a default for this, sort of like: https://github.com/ChameleonCloud/chi-in-a-box/blob/68b2d551c4489841ebe6ff6eaf3746c457fbc124/kolla/defaults.yml#L145

Otherwise this will fail with a "variable not defined" error.

roles/chameleon_usage/defaults/main.yml
```diff
@@ -6,3 +6,23 @@ chameleon_portal_mysql_host: udcc-db01.tacc.utexas.edu
 chameleon_portal_mysql_user: chameleon_ro
 ga_reporting_private_key_json_location: /root/ga-service-account-private-key.json
 google_analytics_view_id: 90514617
+
+usage_report_cron_name: usage_report
```
diurnalist: We may want to make this more specific. What about `user_project_report`?

roles/chameleon_usage/tasks/usage_report.yml
```diff
@@ -0,0 +1,22 @@
+---
```
diurnalist: Suggest we rename this to `user_project_report.yml` (as above suggestion). Or something... but they should be named the same, so it's more obvious what this is, precisely. This role contains several reports, all of which technically are "usage reports", a bit of clarity will greatly help future readers of the code.

diurnalist (approved):
Great stuff! I pushed a little commit to rename the task as well.

PR #54, opened 2020-01-21
[precis] set up cleanup

zhenz:


roles/precis/tasks/main.yml
```diff
@@ -106,6 +106,12 @@
       notify:
         - restart precis endpoint
         - restart precis generator
+    - name: Create service configuration.
+      template:
+        src: "precis_cleanup.conf.j2"
+        mode: 0600
```
diurnalist, line 112: 🔒

roles/precis/tasks/main.yml
```diff
@@ -135,3 +141,19 @@
...
+
+- name: Set up precis cleanup cron job.
+  cron:
+    name: "precis_cleaup"
+    minute: "0"
+    hour: "0"
+    day: "1"
+    month: "*"
+    weekday: "*"
+    job: "/usr/local/sbin/precis_cleanup | /usr/bin/logger -t \"precis_cleanup\""
```
diurnalist: Please import the `cron` role when installing a cron job. It can be done like this:  https://github.com/ChameleonCloud/chi-in-a-box/blob/master/roles/mariadb/tasks/main.yml#L2-L6

That task will ensure a special script is installed on the target box, and store the location to that script as `cron_notification`. Then your cron `job` parameter becomes:

```yaml
job: "/usr/local/sbin/precis_cleanup | /usr/bin/logger -t precis_cleanup && {{ cron_notification }}"
```

The notification script is responsible for reporting a metric (last successful run) to Prometheus, which in turn allows us to detect when jobs start failing for some reason.

diurnalist (approved):
This all looks good in general, see the note about the cron task though :)

Great work!

PR #55, opened 2020-01-23
[precis] configure vhost

zhenz:


roles/apache/templates/vhost.conf.j2
```diff
@@ -1,14 +1,18 @@
 # {{ ansible_managed }}
 # Virtual host for {{ apache_server_name }}
 
+{% if vip_address is not defined %}
```
diurnalist: Nice, I like this. I think a better name for it might be `apache_server_bind_address`, just to match the others; I never really liked how I re-used `external_vip_address`.

playbooks/precis.yml
```diff
@@ -6,3 +6,23 @@
...
+        tasks_from: host
+      vars:
+        vip_address: "{{ precis_vip_address }}"
+        apache_server_priority: 25
+        apache_server_alias: precis_public
+        apache_server_name: "{{ precis_server_hostname }}"
+        apache_ssl_letsencrypt_enabled: no
+        apache_server_https_port: "{{ precis_endpoint_port }}"
+        apache_server_http_port: null
+        apache_server_ssl_certificate_file: "{{ precis_ssl_file_prefix }}_chameleoncloud_org_cert.cer"
```
diurnalist: I generally believe that making these things explicit is a better place to start, especially when the pattern is not yet well-formed (we have 2 examples, maybe we can revisit when we know what TACC / other sites are doing). I suggest `precis_ssl_certificate_file` having the full file name.

diurnalist (approved):
Very nice work!

PR #57, opened 2020-02-07
[hammer] fix quoting issue

zhenz:


diurnalist (approved):
Nice, that's the fix!

PR #58, opened 2020-02-10
[hammers] fix timezone

zhenz:


diurnalist (approved):
So it's enough to simply set TZ inside the container? For some reason I thought CentOS needed more than that. Cool!

zhenz:
> So it's enough to simply set TZ inside the container? For some reason I thought CentOS needed more than that. Cool!

yes. tested on dev@uc.

PR #59, opened 2020-02-20
[nova][neutron] use messagingv2 for message notifications

zhenz:


diurnalist (approved):
👍 , maybe can put a note in there that this is for Precis, so people don't delete it later.

PR #60, opened 2020-02-26
[precis] fixes

zhenz:
1. create service and endpoint
2. create and initialize database
3. register premetheus metrics

### DIFF
--- a/roles/precis/tasks/main.yml
+++ b/roles/precis/tasks/main.yml
@@ -1,4 +1,21 @@
 ---
+- name: Create cep service
+  os_keystone_service:
+    name: cep
+    service_type: cep
+    auth: "{{ openstack_auth }}"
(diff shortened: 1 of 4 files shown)


roles/precis/tasks/main.yml
```diff
@@ -1,4 +1,19 @@
 ---
+- name: Create cep service
+  os_keystone_service:
+    name: cep
```
diurnalist, line 4: You will probably need to have `auth` (or `openstack_auth`, I forget which) here. Otherwise, it will be writing updates to the secondary Keystone server and cause replication errors. There is an `openstack_auth` var already set, I think, which you can use here.

diurnalist (approved):
Nice changes! Looks good to me, left a comment on the auth credentials

PR #62, opened 2020-04-02
Add redfish-monitor

jakecoll:
Redfish Monitor code here: https://github.com/ChameleonCloud/redfish-monitor

### DIFF
--- a/roles/redfish_monitor/tasks/main.yml
+++ b/roles/redfish_monitor/tasks/main.yml
@@ -21,20 +21,16 @@
     mode: 0600
     dest: "{{ redfish_monitor_env_file }}"
 
-- name: Start redfish-monitor docker container
-  docker_container:
-    name: redfish_monitor
-    auto_remove: yes
-    image: "{{ redfish_monitor_docker_image }}"
-    network_mode: host
-    env_file: "{{ redfish_monitor_env_file }}"
-    volumes:
-      - "{{ redfish_monitor_log_folder }}:{{ redfish_monitor_log_folder }}"
-    command:
-        - "redfish-monitor"
-        - "-u"
-        - "$REDFISH_USER"
-        - "-p"
-        - "$REDFISH_PASSWORD"
-        - "--minutes"
-        - "{{ redfish_monitor_minute_interval }}"
+- name: Create redfish_monitor services.
+  template:
+    src: "redfish_monitor.service.j2"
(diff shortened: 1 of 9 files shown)


roles/redfish_monitor/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+    path: "{{ redfish_monitor_log_folder }}"
+    state: directory
+
+- name: Configure redfish-monitor env file
+  template:
+    src: redfish-monitor.env.j2
+    mode: 0600
+    dest: "{{ redfish_monitor_env_file }}"
+
+- name: Start redfish-monitor docker container
```
diurnalist: We normally provision these things as systemd services - there are pros and cons. What were your reasons for using the `docker_container` mechanism here?

roles/redfish_monitor/templates/redfish-monitor.env.j2
```diff
@@ -0,0 +1,12 @@
+OS_AUTH_STRATEGY=keystone
+OS_AUTH_URL={{ keystone_internal_url }}/v3
```
diurnalist, line 2: Make sure that internal_url isn't suffixed with /v3, this has been a problem in the past I think.

roles/redfish_monitor/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+  docker_container:
+    name: redfish_monitor
+    auto_remove: yes
+    image: "{{ redfish_monitor_docker_image }}"
+    network_mode: host
+    env_file: "{{ redfish_monitor_env_file }}"
+    volumes:
+      - "{{ redfish_monitor_log_folder }}:/var/log"
```
diurnalist: `/var/log/` is kind of generic: is it possible that other logs would accidentally be written there, that we don't want? Like, syslog normally goes there, or /var/log/messages, which we presumably don't want in particular. Though, it might not be an issue for application containers, which don't run with an init system.

roles/redfish_monitor/tasks/main.yml
jakecoll: The asynchronous code requires python3 which we don't have configured on admin01.

diurnalist: In the prometheus role we set up systemd services that call Docker: https://github.com/ChameleonCloud/chi-in-a-box/blob/master/roles/prometheus/templates/systemd.service.j2

PR #63, opened 2020-04-02
Default to Train release

diurnalist:
All of our clouds are now on Train, so it makes sense to default, lest we accidentally revert to Rocky in the future.

### DIFF
--- a/kolla/node_custom_config/nova.conf
+++ b/kolla/node_custom_config/nova.conf
@@ -6,7 +6,7 @@ vif_plugging_timeout = 0
 max_concurrent_builds = 0
 
 [compute]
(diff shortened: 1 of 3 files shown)


kolla/node_custom_config/nova.conf
```diff
@@ -6,7 +6,7 @@ vif_plugging_timeout = 0
 max_concurrent_builds = 0
 
 [compute]
-# https://docs.openstack.org/ironic/rocky/install/configure-compute.html
+# https://docs.openstack.org/ironic/train/install/configure-compute.html
```
codyhammock, line 9: Should this be parameterized with `{{ ironic_tag }}`?

PR #68, opened 2020-04-14
[prometheus] add precis alert

zhenz:


roles/chameleon_prometheus/files/rules/control-rules.yml
```diff
@@ -74,3 +74,16 @@ groups:
       summary: "OpenStack {{ $labels.service | title }} service is down"
       description: "The *{{ $labels.service | title }}* service on {{ $labels.hostname }} is not responding."
       runbook: https://github.com/ChameleonCloud/chi-in-a-box/wiki/%5BRunbook%5D-OpenStackAPIDown
+
+  - alert: ParsedPrecisEventsLow
```
diurnalist: Suuuper nitpick, but the convention is to put the "big noun" first, as a sort of namespacing convention. In this case the "big noun" would be Precis.

Suggestion: `PrecisParsedEventsLow`

roles/chameleon_prometheus/files/rules/control-rules.yml
```diff
@@ -74,3 +74,16 @@ groups:
...
+    expr: |
+      rate(precis_events_total{service!="glance",type="parsed"}[3d]) == 0
```
diurnalist: I believe it will be safer long term to make this a `sum()` aggregation:

```prometheus
sum(rate(precis_events_total{service!="glance",type="parsed"}[3d])) by (service) == 0
```

The reason is that it may be the case down the road that the exporter moves to a new host or something. In that case, the `instance` label on this metric (which is automatically added by Prometheus) will be different. I'm not 100% sure of how the alert engine deals with this, but it could mean that an alert erroneously triggers because for _some_ label combination (the label combination that effectively no longer exists), the rate was 0. `sum()` is useful for compressing/removing labels that are not relevant to the alert.

This makes the alert a bit annoying to write because of the Glance special case. Duplicating is fine:

```prometheus
sum(rate(precis_events_total{service!="glance",type="parsed"}[3d])) by (service) == 0
or
sum(rate(precis_events_total{service="glance",type="parsed"}[1w])) by (service) == 0
```

An alternative is to create a new _expression_, which is basically a derived metric that Prometheus calculates for you every so often (I think every 60s by default?)

[Naming rules](https://prometheus.io/docs/practices/rules/) for these are quite complicated. I think in our case it makes sense to do something like:

```yaml
- record: service_type:precis_events_total:rate1w
  expr: sum(rate(precis_events_total{type="parsed"}[1w])) by (service, type)

- alert: PrecisParsedEventsLow
  expr: service_type:precis_events_total:rate1w{type="parsed"} == 0
  for: 1d
```

I would say, for simplicity, perhaps let's just always look at the rate over the last week. It will be a bit slower to pick up on issues, but it will probably be less brittle over time, and easier to maintain.

roles/precis/defaults/main.yml
```diff
@@ -16,6 +16,7 @@ precis_rabbitmq_listener_bindings:
   - neutron
   - nova
   - openstack
+  - glance
```
diurnalist, line 19: Nice catch :) - I guess the alerts are already useful

diurnalist (approved):
Approved, with some comments on things we can do to make it a bit simpler

⭐

roles/chameleon_prometheus/files/rules/control-rules.yml
```diff
@@ -74,3 +74,16 @@ groups:
...
+  
+  - alert: PrecisParsedEventsLow
+    expr: service_type:precis_events_total:rate1w == 0
+    for: 1d
+    labels:
+      severity: warning
+    annotations:
+      summary: "Experiment Precis event listener stops parsing OpenStack *{{ $labels.service }}* service notification messages"
+      description: "The Experiment Precis parsed no *{{ $labels.service }}* events for the last 1 week."
+      runbook: https://github.com/ChameleonCloud/chi-in-a-box/wiki/%5BRunbook%5D-PrecisParsedEventsLow
```
diurnalist, line 89: This is a fantastic runbook! And thank you for including it with the alert :)

diurnalist (approved):
[nailedit](https://media2.giphy.com/media/9JlLx5uYEQTOvnOGQy/giphy.gif)

PR #71, opened 2020-04-28
add allocations playbook

zhenz:


roles/allocations/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+
+- name: Configure allocations services
+  block:
+    - name: Create Docker network.
+      docker_network:
+        name: "{{ allocations_docker_network }}"
+        ipam_config:
+          - subnet: "{{ allocations_docker_network_subnet }}"
+        state: present
+      when: allocations_docker_network != "host"
```
diurnalist: Good catch

roles/allocations/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+    - name: Set bind address.
```
diurnalist: I recently added a Galaxy role that makes this whole process simpler. We set up all of our Docker services in more or less exactly the same manner--it made sense to generalize the pattern.

Here is an example of how it's done:

https://github.com/ChameleonCloud/chi-in-a-box/blob/421a8f0ed5c13468c677fe16f2cc7f7a57ebfb0d/roles/keycloak/tasks/main.yml#L19-L32

It simplifies the task and means you don't have to include the systemd service files at all! Want to be a beta tester? I've already used it for Keycloak but this is a good chance to try it out in a different context.

roles/allocations/defaults/main.yml
```diff
@@ -0,0 +1,14 @@
...
+# Corresponds to the default user 1001 used in s2i builds
+allocations_user: 1001
+
+allocations_balance_service_docker_image: docker.chameleoncloud.org/allocations/balance-service:latest
+allocaitons_balance_service_name: allocations-balance-service
+allocations_balance_service_port: 8915
+
+redis_host: chi.uc.chameleoncloud.org
+# comma sperated auth users
+allocations_auth_users: 97d6f370dbc5451da5521633eeda5ad6
```
diurnalist: Might be easier to have this be a YAML array, and then use the `join` filter when rendering it out on the template. In general I think it's easier to keep lists as lists in YAML

```jinja
comma_separated_list = "{{ true_list | join(",") }}"
```

roles/allocations/templates/allocations_balance_service.conf.j2
```diff
@@ -0,0 +1,10 @@
+[DEFAULT]
+debug: False
+
+[serviceaccount]
+auth_url: {{ keystone_public_url }}/v3
+auth_users: {{ allocations_auth_users }}
+
+[redis]
+host: {{ redis_host }}
+port: 6379
```
diurnalist: I think `{{ redis_port }}` might be already a built-in variable in Kolla-Ansible we could use.

roles/allocations/templates/allocations_balance_service.conf.j2
```diff
@@ -0,0 +1,10 @@
+[DEFAULT]
+debug: False
```
diurnalist, line 2: Is this syntax right? For some reason I thought it needed `=` instead of `:`. Can it do both? Am I crazy?? :D

diurnalist (changes requested):
- Try out that new Ansible Galaxy role, I think it can trim the required code significantly
- Maybe the list can be expressed differently

Otherwise this looks great! Nice work.

stonecold

roles/allocations/templates/allocations_balance_service.conf.j2
zhenz: I think you asked this before:) `:` works.

diurnalist: I am going crazy then, good to know! 🤪

roles/allocations/tasks/main.yml
```diff
@@ -0,0 +1,47 @@
...
+    src: "allocations_balance_service.conf.j2"
+    mode: 0600
+    dest: "{{ allocations_config_path }}/service.conf"
+    owner: "{{ allocations_user }}"
+
+- name: Configure allocations services
+  include_role:
+    name: chameleon.docker_service
+  vars:
+    service_name: "{{ allocaitons_balance_service_name }}"
```
diurnalist:
```suggestion
    service_name: "{{ allocations_balance_service_name }}"
```

roles/allocations/tasks/main.yml
```diff
@@ -0,0 +1,47 @@
...
+    service_bind_address: "{{ ansible_facts[network_interface]['ipv4']['address'] }}"
```
diurnalist: You can simply do `{{ api_interface_address }}`, I think that is the easiest.

roles/allocations/tasks/main.yml
```diff
@@ -0,0 +1,47 @@
...
+    docker_image: "{{ allocations_balance_service_docker_image }}"
+    docker_network: "{{ allocations_docker_network }}"
+    docker_mounts: 
+      - type: bind
+        src: "{{ allocations_config_path }}/service.conf"
+        dst: "/etc/allocations/balance-service.conf"
+    docker_ports: 
+      - "{{ allocations_balance_service_port }}:8080/tcp"
```
diurnalist: To avoid binding the port on all host interfaces, restrict to just the bind address:

```suggestion
      - "{{ api_interface_address }}:{{ allocations_balance_service_port }}:8080/tcp"
```

roles/allocations/tasks/main.yml
```diff
@@ -0,0 +1,47 @@
...
+
+- name: Enable balance service on boot.
```
diurnalist: I think the `docker_service` role will do this for you.

roles/allocations/tasks/main.yml
```diff
@@ -0,0 +1,47 @@
...
+- name: Pull allocations service container images.
+  docker_image:
+    source: pull
+    name: "{{ item }}"
+    state: present
+    force_source: yes
+  loop:
+    - "{{ allocations_balance_service_docker_image }}"
+  tags:
+    - pull
```
diurnalist, line 13: This should `notify` the service to restart

roles/allocations/tasks/main.yml
```diff
@@ -0,0 +1,47 @@
...
+      - subnet: "{{ allocations_docker_network_subnet }}"
+    state: present
+  when: allocations_docker_network != "host"
+
+- name: Create init configuration.
+  template:
+    src: "allocations_balance_service.conf.j2"
+    mode: 0600
+    dest: "{{ allocations_config_path }}/service.conf"
+    owner: "{{ allocations_user }}"
```
diurnalist, line 28: This should also `notify` the service

roles/allocations/handlers/main.yml
```diff
@@ -0,0 +1,13 @@
+---
+# This needs to run before the other handlers.
+# "Handlers are run in the order they are listed, not in the order that they
+# are notified."[1]
+# [1]: https://docs.ansible.com/ansible/latest/reference_appendices/glossary.html#notify
+- name: "reload systemd daemon"
```
diurnalist: I don't think this is needed anymore; the `docker_service` role can cover it. The other handler though, to restart the service itself, is still useful, so when configuration changes, it triggers an automatic restart of the service to pick up the changes.

diurnalist (approved):
Approach looks good, I left some suggestions we should apply before release (binding on internal address and simplifying how we get that address are the most important.)

nailedit

diurnalist (approved):
Looks great! Thanks for all the changes :)

PR #72, opened 2020-05-14
Add support for Keystone federation

diurnalist:
This adds support for Keystone federation in CHI-in-a-Box. By default, if Portal websso is enabled, that will still be the default, and you have to opt in via a special Horizon endpoint (`/auth/new-login-flow`) in order to get the new UI, which supports logging in either via Portal or via Keycloak.

### DIFF
--- a/roles/keycloak/tasks/import_realm.yml
+++ b/roles/keycloak/tasks/import_realm.yml
@@ -1,6 +1,0 @@
----
-- name: Import realm
-  command: >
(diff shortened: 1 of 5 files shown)


roles/keycloak/tasks/import_realm.yml
```diff
@@ -0,0 +1,6 @@
+---
```
diurnalist: Note to self: remove this

PR #73, opened 2020-05-22
[precis] use docker_service and haproxy

zhenz:


roles/precis/defaults/main.yml
```diff
@@ -49,3 +49,20 @@ precis_generator_service_name: experiment-precis-generator
 precis_generator_prometheus_port: 8912
 
 precis_cleanup_docker_image: docker.chameleoncloud.org/experiment-precis/cleanup:latest
+
+precis_services:
+  precis:
+    group: precis
+    enabled: "yes"
```
diurnalist: can be derived from `{{ enable_precis }}`, which I think exists

diurnalist (approved):
Very cool! And you are using the docker_service role now - great!

:shipit:

PR #81, opened 2020-07-16
[keycloak_tas_sync] add portal env variables

zhenz:


roles/keycloak_tas_sync/defaults/main.yml
```diff
@@ -18,9 +18,16 @@ keycloak_tas_sync_tasks:
...
+        PORTAL_MYSQL_USER: "{{ portal_mysql_user }}"
+        PORTAL_MYSQL_PASSWORD: "{{ portal_mysql_password }}"
+        PORTAL_MYSQL_DATABASE: "{{ portal_mysql_database }}"
 
 keycloak_tas_sync_ldap_bind_uid: cctest
 keycloak_tas_sync_tas_client_key: tasclient
 keycloak_tas_sync_keycloak_client_id: user-group-import
 keycloak_tas_sync_provider_alias: tacc
 keycloak_tas_sync_provider_sub: 6c3f8474-e0f7-49f7-9ddd-90c1b899daab
+portal_mysql_host: udcc-db01.tacc.utexas.edu
```
diurnalist: This is the third place we now have this host listed. I think we should take these variables `portal_*` and lift them into kolla/defaults.yml

diurnalist (approved):
:shipit: ! But, I suggest moving the variables up to defaults. We'll also need to update site-configs w/ the password if it isn't already there.

PR #83, opened 2020-07-21
[allocations] Updates for Balance Service Usage Enforcment

jakecoll:
The allocations service now handles usage enforcement via the
external service filter in Blazar. Edits here include adding
usage enforcement configs and an allowed_auth_urls variable to
enable balance service to authenticate against multiple keystone
urls.


diurnalist (approved):
Very clean, very nice.

PR #87, opened 2020-08-06
[blazar] Add new usage enforcement defaults

jakecoll:


kolla/defaults.yml
```diff
@@ -30,13 +30,15 @@ blazar_project_enforcement_id: charge_code
 # When to send lease notification email
 blazar_minutes_before_end_lease: 2880
 blazar_default_max_lease_duration: 604800
+blazar_default_reservation_extension_window: 172800
 blazar_usage_default_allocated: 20000.0
 # Redis currently binds on public IP, use public FQDN
 blazar_usage_db_host: "{{ kolla_external_fqdn }}"
 blazar_usage_project_exemptions:
```
diurnalist: These don't seem to be used anywhere. I think it needs to be in the `[enforcement]` config section.

diurnalist (changes requested):
I'd actually say that the list of exempt projects should be an empty list by default, and they should be overridden in the site-configs. Each site will have a different list of project IDs potentially. It just so happens that prod UC/TACC share project IDs at the moment; this need not be the case in the future. Also think of the fact that CHI-in-a-Box will be used by other deployers who don't have any of these projects configured. What do you think?

kolla/defaults.yml
jakecoll: It is in enforcement:

```python
reservation_extension_window = {{ blazar_default_reservation_extension_window }}
```

This is what allows users to extend within 48 hours.

jakecoll:
> I'd actually say that the list of exempt projects should be an empty list by default, and they should be overridden in the site-configs. Each site will have a different list of project IDs potentially. It just so happens that prod UC/TACC share project IDs at the moment; this need not be the case in the future. Also think of the fact that CHI-in-a-Box will be used by other deployers who don't have any of these projects configured. What do you think?

I can update.

kolla/defaults.yml
diurnalist: I was referring to the project whitelist -- where is that used in the updated Blazar config?

diurnalist:
OK, let's move the whitelist to the site-configs then. It should still have a default set up in defaults.yml here so that it doesn't break if you don't override it to anything.

kolla/defaults.yml
jakecoll: Right right. I've set default to an empty list and updated site-configs.

PR #88, opened 2020-08-21
[jupyterhub] Convert to services + add Trovi

diurnalist:
This adds Trovi configuration for JupyterHub, and also converts
JupyterHub to use the new services framework. When this is rolled out,
it will be served by HAProxy.

The plan would be to roll this out, and then do a DNS record switchover,
at which point Apache will stop serving the app, and HAProxy takes over.


cc-ansible
```diff
@@ -220,7 +220,9 @@ if [[ "$CHECK_UPDATES" == "yes" || "$FORCE_UPDATES" == "yes" ]]; then
         ! diff -q >/dev/null \
           $kolla_ansible_gitref \
           <(cd $kolla_ansible_egglink; git fetch; git show-ref -s -d origin/$kolla_ansible_checkout); then
+    pushd $VIRTUALENV
```
diurnalist, line 223: Pip apparently installs egg source files relative to the current directory, somehow I didn't catch this before (or maybe it changed.)

PR #92, opened 2021-01-06
Cross site backup

hlbfrancois:


roles/chameleon_mariadb/tasks/main.yml
```diff
@@ -19,6 +19,37 @@
     dest: "{{ mariadb_backup_cron_script }}"
   when: inventory_hostname_short == "dev"
 
+- name: Create directory to mount to aws-cli docker container for copying config and backup files to container
+  file:
+    path: /etc/awsconf
+    state: directory
+  when: inventory_hostname_short == "admin01" or
```
diurnalist: A role should not need to understand which hosts it is running on. The inventory decides how to map roles to hosts or groups of hosts. You can imagine that this approach becomes difficult to scale to TACC=>UC, many of the templates would need duplicating and tasks too. However, almost all of what you have is very close to being entirely generic!

What I recommend:
- Create a new variable to hold the value of the backup destination (the Ceph S3 endpoint). Use this variable when templating the backup script.
- Consider setting aws credentials as environment variables to the container (if possible), this will remove the need for that additional file just to hold the creds
- Rename `tacc_ceph_backup_access_key` to just `mariadb_ceph_backup_access_key`. Typically variables are prefixed w/ the name of the role they are tied to. This also allows each site to set a different value for this in site-configs.

diurnalist (changes requested):
This is very close! Does the Ceph cluster need to be configured in some special way first? Might be good to call that out in the PR / commit.

roles/chameleon_mariadb/tasks/main.yml
hlbfrancois: Thanks Jason, I've made all the updates.

roles/chameleon_mariadb/templates/mariadb.j2
```diff
@@ -45,14 +45,22 @@ mysqldump --defaults-extra-file=$TMPFILE --opt --flush-logs --single-transaction
...
 
-# Remove temporary file
+# Remove temporary files
 rm -f $TMPFILE
+rm ${S3DIR}/${LATEST_FILE}
 
 
+if [ $? -eq 0 ] ; then
+    cleanup
+    touch /tmp/mysqlbackup_success
```
diurnalist: Why is this necessary?

roles/chameleon_mariadb/templates/mariadb.j2
```diff
@@ -45,14 +45,22 @@ mysqldump --defaults-extra-file=$TMPFILE --opt --flush-logs --single-transaction
...
-fi
+# AWS S3  cross site backup
+# Copy latest backup to /etc/awsconf to mount to container
+S3DIR="{{ mariadb_backup_s3_conf_dir }}"
+LATEST_FILE=`ls -rt ${DIR} | grep mysql_backup | tail -1`
+cp $DIR/$LATEST_FILE $S3DIR
 
-rsync -vrlptz ${DIR} "{{ mariadb_backup_destination_host }}":"{{ mariadb_backup_destination_dir }}" 2>&1 | /usr/bin/logger -t "{{ mariadb_backup_log_key }}" && ssh "{{ mariadb_backup_destination_host }}" 'find {{ mariadb_backup_destination_dir }}* -maxdepth 2 -mtime "{{ mariadb_backup_retention }}" -exec rm {} \;'
+# Create bucket and copy backup file, cleanup file
+docker run -e AWS_ACCESS_KEY_ID="{{ mariadb_backup_s3_key_id }}" -e AWS_SECRET_ACCESS_KEY="{{ mariadb_backup_s3_key }}" --rm -it -v ${S3DIR}:/root/.aws amazon/aws-cli --endpoint="{{ mariadb_backup_s3_endpoint }}" --region=us-east-1 s3 mb s3://"{{ mariadb_backup_container_name }}"
```
diurnalist: Does `--region` need to be us-east-1? Does it just need to be set to a value? If so, maybe set it to `dummy` or something so it's clear it's not really being used.

diurnalist (approved):
LGTM! 🏆

roles/chameleon_mariadb/templates/mariadb.j2
hlbfrancois: I thought it did....but apparently it likes Dummy :)

roles/chameleon_mariadb/templates/mariadb.j2
hlbfrancois: I don't think it is, it was already there. From what I read in man pages, cleanup doesn't do much for us here. And the success touch is redundant since we're logging results, right? I've removed it.

roles/chameleon_mariadb/templates/mariadb.j2
hlbfrancois: So it likes dummy if I just do an 'ls' on the bucket, but if I try to create a bucket it says "The specified location-constraint is not valid", so I'll revert.

PR #95, opened 2021-02-08
[chameleon_usage] add kvm usage report

zhenz:


roles/chameleon_usage/tasks/config.yml
```diff
@@ -17,3 +17,4 @@
   vars:
     mysql_host: "{{ chameleon_usage_mysql_host }}"
     root_password: "{{ database_password }}"
+    kvm_host: "{{ kvm_mysql_host }}"
```
diurnalist: Without a default or something, this will fail the role at UC. Can you make it so the KVM configuration is optional and will only run if this variable (or something like `enable_usage_kvm` is True, which is how we usually toggle these things on/off.)

diurnalist (changes requested):

zhenz:
@diurnalist only updated chi report for this PR. As we discussed, the KVM report will be enabled on the KVM site only and changes should be made to the kvm branch. Will create a separate PR.

roles/chameleon_usage/templates/chameleon-usage-extract-chi-data.j2
```diff
@@ -4,4 +4,4 @@ docker run --rm --net=host \
     -v "{{ backup_dir }}:{{ backup_dir }}" \
     -v "{{ chameleon_usage_config_dir }}/my.cnf:/etc/mysql/my.cnf" \
     {{ chameleon_usage_docker_image }} \
-    extract_data --site {{chameleon_site_name}} --backup-location {{ backup_dir }} --mysql-conf /etc/mysql/my.cnf 2>&1
+    extract_chi_data --site {{chameleon_site_name}} --backup-location {{ backup_dir }} --mysql-conf /etc/mysql/my.cnf 2>&1
```
diurnalist: You could reuse this whole thing if the name of the function to run was also parameterized. What if we did that? Then you wouldn't have to create the separate KVM merge; you could simply set `usage_extract_command: extract_chi_data` versus `usage_extract_command: extract_kvm_data`.

diurnalist (approved):
Approved, but consider that we could add 1 variable to make this generic for KVM and CHI collection; you've already done all the other work of making the process so similar it's just the difference b/w a script name :)

PR #97, opened 2021-02-12
Load list of tags to skip from file

msherman64:
kolla-ansible supports the --skip-tags option

This adds a simple list of tags to a file. Uncomment lines to skip.


kolla-skip-tags
```diff
@@ -0,0 +1,72 @@
+#!/bin/sh
```
diurnalist: Doesn't need this does it?

cc-ansible
```diff
@@ -325,4 +325,11 @@ if [[ "${POSARGS[*]}" =~ post-deploy ]]; then
 fi
 kolla_args+=("${POSARGS[@]}")
 
+#load tags to skip
+ANSIBLE_SKIP_TAGS="$(grep -v "^#" ./kolla-skip-tags | paste -sd ',')"
+if [[ "${ANSIBLE_SKIP_TAGS}" ]]; then
```
diurnalist: Huh, I didn't know it could be that simple. Is there a difference b/w this and `-n` ?

cc-ansible
```diff
@@ -325,4 +325,11 @@ if [[ "${POSARGS[*]}" =~ post-deploy ]]; then
...
+  kolla_args+=("--skip-tags")
+  kolla_args+=("${ANSIBLE_SKIP_TAGS[@]}")
```
diurnalist: Do you need the `[@]`?

diurnalist (approved):
Looks good, I like that they are in order!

kolla-skip-tags
msherman64: This was just to enable syntax highlighting in general. It's getting evaluated by bash / gnu functions, so it seemed reasonable to include.

msherman64:
The list of tags was generated by running the raw ansible-playbook command with `--list-tags`, then manually parsing it.
It might be good to have some kind of CI process generate that list, so it stays up to date.

cc-ansible
msherman64: I need to check, that's left over from when it was getting interpreted as an array.

cc-ansible
msherman64: With no arguments, `[[ expr ]]` tests for "is null", where `-n` tests for "is length nonzero"
From what I can find: https://unix.stackexchange.com/a/280432 they result in the same behavior, but `-n` is semantically clearer.

PR #98, opened 2021-02-12
Bugfix/centos8-compat

msherman64:
Upstream has introduced a number of changes for centos8, which require the use of centos8 containers on a centos8 host.

This draft PR is holding changes I add to get `./cc-ansible deploy` to run successfully on a centos8 host.

### DIFF
--- a/kolla/node_custom_config/horizon/openrc.sh.template
+++ b/kolla/node_custom_config/horizon/openrc.sh.template
@@ -14,7 +14,7 @@ export OS_PASSWORD=$OS_PASSWORD_INPUT
 {% endraw %}
 export OS_IDENTITY_PROVIDER="{{ identity_provider_name }}"
 export OS_DISCOVERY_ENDPOINT="{{ identity_provider_url }}/.well-known/openid-configuration"
-export OS_CLIENT_ID="{{ keystone_idp_client_id }}"
+export OS_CLIENT_ID="{{ keystone_idp_client_id|default(none) }}"
 export OS_ACCESS_TOKEN_TYPE="access_token"
 {% raw -%}
 export OS_CLIENT_SECRET="none"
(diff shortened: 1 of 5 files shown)


kolla/node_custom_config/horizon/openrc.sh.template
```diff
@@ -14,7 +14,7 @@ export OS_PASSWORD=$OS_PASSWORD_INPUT
 {% endraw %}
 export OS_IDENTITY_PROVIDER="{{ identity_provider_name }}"
 export OS_DISCOVERY_ENDPOINT="{{ identity_provider_url }}/.well-known/openid-configuration"
-export OS_CLIENT_ID="{{ keystone_idp_client_id }}"
+export OS_CLIENT_ID="{{ keystone_idp_client_id|default(none) }}"
```
diurnalist, line 17: Should we just set this to something in defaults.yml instead of defaulting everywhere it is used?
Also, what about setting `enable_keystone_federation: no` in defaults?

diurnalist (approved):
I'm cool with these changes--wdyt about changing the default settings to spin up the site w/o federation enabled? We can then document how to enable federation, as it's a second step and involves asking us for a client to be created.

kolla/node_custom_config/horizon/openrc.sh.template
msherman64: Currently `keystone_idp_client_id` MUST be set to something, even null.
Even in the case with:
`enable_keystone_federation: no` and `enable_keystone_federation_openid: no`

To resolve this, I think it makes sense to look at the desired behavior of the template.
https://docs.ansible.com/ansible/2.8/user_guide/playbooks_filters.html#omitting-parameters documents some options.

We can choose between:

1. Fail on an undefined variable
2. set a hardcoded default: `default(value)`
3. set a configurable default, or chain thereof: `default(defaults.idp_client_id) | defaults(none)`
4. omit the templated line `default(omit)`

PR #101, opened 2021-05-03
Deprecated: Deploy KVM flavor of CHI

diurnalist:


msherman64:
It appears that the only change needed to "turn on" KVM support is `nova_compute_virt_type: kvm`
And possibly our default values for quotas, and how they interact between vm and baremetal.

msherman64:
`nova_compute_virt_type: kvm` is now present in our supported branches of chi-in-a-box, and I've verified that usage of kvm instances "works" with the other defaults.

However, the user workflow if both KVM and baremetal are enabled is a bit convoluted.

Closing for now, but we may want to dig through this branch for nova tuning options later.

PR #106, opened 2021-06-30
Add support for deploying Zun for edge-flavored

diurnalist:
This add support for deploying edge services, including Zun. It is intended primarily for adding support for CHI@Edge, but Zun could also be deployed as part of a separate cloud already supporting KVM or Ironic deployments.

### Allowing hosts/groups to override globals

Because edge devices are radically different than the management nodes running the services, and currently we still use chi-in-a-box to provision the edge devices, some things had to change. Mainly, "--globals" no longer cuts the mustard because it applies those overrides to every host. We needed a way to allow hosts to override these values. To fix this, cc-ansible will now create a tmpdir and assemble it by doing a few things:

1. Merge defaults.yml from chi-in-a-box with an optional "defaults.yml" defined in the site-config. `yq` is used for this and is now a requirement for chi-in-a-box. It will be installed into the virtualenv automatically.
2. Copy node_custom_config from site config, if defined. Any files in here will *overwrite* one of the same name defined in chi-in-a-box's kolla/node_custom_config. The main use-case for providing this in a site config is for ceph keys, which kolla-ansible expects to find in node_custom_config.
3. Copy globals.yml if defined. This has the same behavior as before; it overrides any variables defined in defaults.yml or in host/group_vars

**This should be backwards-compatible.** However, it is probably good to transition site-configs from globals.yml to defaults.yml, to allow them to override per-host/per-group settings if desired.

### Adding l2population ml2 plugin by default

I believe this does not have any effect if tunnels are not being used. But, it is required for VxLAN to work, and is non-obvious to have to add if using Zun or Nova+KVM.

## New stuff

Zun and Kuryr are now supported. To configure a site for an edge-flavored deployment, update defaults.yml in the site-config:

```yaml
# Required
enable_etcd: yes
enable_kuryr: yes
enable_zun: yes
# Optional
enable_nova: no
enable_ironic: no
```


diurnalist:
(at this point we should set aside time to rewrite cc-ansible in a real programming language ;))

msherman64:
LGTM

msherman64 (approved):

codyhammock (approved):
LGTM. I think I just want to make sure I know exactly what to do with node_custom_config when bringing this to production.

PR #107, opened 2021-07-16
add osg support

zhenz:


kolla/node_custom_config/nova/policy.json
```diff
@@ -1,3 +1,7 @@
 {
-  "os_compute_api:os-extended-server-attributes": "rule:admin_or_owner"
+  "os_compute_api:os-extended-server-attributes": "rule:admin_or_owner",
+  "compute:servers:create:requested_destination": "role:osg or rule:admin",
```
diurnalist, line 2: nitpick, can you put these in alphabetical order please?

roles/osg/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
+---
+- name: Create osg role
+  os_keystone_role:
```
diurnalist, line 3: Please wrap this in a `kolla_toolbox` invocation, these keystone modules are really annoying to use reliably on the host system; executing w/in the kolla_toolbox is by far the most reliable method I've found.

roles/osg/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
...
+    auth: "{{ openstack_auth }}"
+    name: osg
+    
+- name: "Add service user to osg project as osg"
+  os_user_role:
```
diurnalist: Same for this one

roles/osg/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
...
+    auth: "{{ openstack_auth }}"
+    user: "{{ osg_service_username }}"
+    role: osg
+    project: "{{ osg_project_id }}"
```
diurnalist: Does it work to specify a project name/domain name here? That is probably easier to manage, as we only have to ensure the names are the same across sites for the play to be generic.

roles/osg/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
...
+- name: Create openstack env file
+  template:
+    src: "openstack.env.j2"
+    mode: 0640
+    dest: "{{ osg_config_path }}/openstack.env"
+
+- name: Configure osg deployer docker bash wrapper
+  template:
+    src: osg-deployer.j2
+    mode: a+x
```
diurnalist: I'm not sure, but I thought things in /sbin don't have executable by all set. It would be more like 0750. But maybe double-check that.

roles/osg/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
...
+    dest: /usr/local/sbin/osg-deployer
+
+- name: Set up osg deployer periodic task
+  include_role:
+    name: chameleon.periodic_task
```
diurnalist, line 60: 👍🏻

roles/osg/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
...
+  vars:
+    task_name: "osg_deployer"
+    task_command: "/usr/local/sbin/osg-deployer"
+    task_calendar: "*:0/30"
+  when: enable_osg | bool
+  
+- name: Check if osg timer exists
+  stat: path=/etc/init.d/osg_deployer
```
diurnalist, line 68: I don't understand what this is doing -- what is this and the next task for?

roles/osg/templates/osg-deployer.j2
```diff
@@ -0,0 +1,7 @@
+#! /bin/sh
+
+docker run --rm --net=host \
```
diurnalist: Does this need "--net=host"? I think it's probably OK if it does, just wondering.

roles/osg/templates/osg.env.j2
```diff
@@ -0,0 +1,5 @@
+TOKEN={{ osg_token }}
+GLIDEIN_Site=Chameleon
+GLIDEIN_ResourceName={{ osg_site_resource_name }}
+GLIDEIN_Start_Extra=True
+OSG_SQUID_LOCATION=http://127.0.0.1:3128
```
diurnalist, line 5: Ah, this is the reason for "--net=host", I take it? OK. Does the OSG Squid thing run in a container as well? If so, we can connect them together and then this problem goes away.

diurnalist (approved):
I think in general this looks great, left some suggestions and a question about the Squid proxy.

🏆

roles/osg/tasks/main.yml
zhenz: if enable_osg is false, check if osg timer exists and is started. if yes, stop the timer.

PR #112, opened 2021-08-06
Zun host policy update

Mark-Powers:
Allows
```
openstack appcontainer host list
openstack appcontainer host show <host>
```

### DIFF
--- a/kolla/node_custom_config/zun/policy.yaml
+++ b/kolla/node_custom_config/zun/policy.yaml
@@ -2,3 +2,5 @@
 container:create:runtime: rule:admin_or_owner
 container:create:requested_destination: rule:admin_or_owner
 container:delete_force: rule:admin_or_owner
+host:get: "@"
+host:get_all: "@"


kolla/node_custom_config/zun/policy.yaml
```diff
@@ -2,3 +2,5 @@
 container:create:runtime: rule:admin_or_owner
 container:create:requested_destination: rule:admin_or_owner
 container:delete_force: rule:admin_or_owner
+host:get: "@"
```
diurnalist, line 5: Bit of a weird syntax, but OK! Maybe should have a comment "@ = allow all" above?

diurnalist (approved):
Thanks!

kolla/node_custom_config/zun/policy.yaml
Mark-Powers: Yes, there are a few ways to allow all, the other two looked a bit too much like "allow none" for me. They are:
[] and ""

https://docs.openstack.org/oslo.policy/latest/admin/policy-yaml-file.html#syntax

PR #121, opened 2021-09-17
Bugfix/shell robustness

msherman64:
Handle common error cases in shell scripts

`set -o errexit` can be bypassed by adding `|| true`
`set -o pipefail` will exit if any part of a pipe fails, rather than just the last. Can be bypassed as with errexit.
`set -o nounset` will error if an undeclared variable is used. To test an unset variable, use `"${varname:-}"`


cc-ansible
```diff
@@ -357,7 +368,7 @@ kolla_args+=(--extra cc_ansible_site_dir="$CONFIG_DIR")
 # Override python interpreter to point to virtualenv (but only if we're
 # not performing the bootstrap, which is responsible for setting up the
 # virtualenv in the first place!)
-if [[ ! "${POSARGS[@]}" =~ bootstrap-servers ]]; then
+if [[ ! "${POSARGS[*]}" =~ bootstrap-servers ]]; then
```
diurnalist, line 371: Good catch, I didn't think about the fact that @ isn't the right thing here

diurnalist (approved):
LGTM! I usually indeed to `set -e -u -o pipefail` at the top of things, not sure why it's not already here. Good to have! 🚢

PR #122, opened 2021-09-17
add basic exercise of chi-in-a-box script

msherman64:
Very basic testing, just make sure that init and genconfig run properly on the ephemeral GH actions VMs with the default site-config


diurnalist (approved):
Very nice

PR #124, opened 2021-09-17
Enhancement/standard venv paths

msherman64:
Instead of sourcing the virtualenv inside the script, manually set the standard variables
Python uses `VIRTUAL_ENV` and `PATH` for detecting whether its in a virtualenv or not, and to run the correct binaries.

In addition, change the default venv path to `.venv` to match python defaults.

Benefits:
- will reuse an existing venv if already sourced
- compatibility with more tools, possible easier migration to poetry.

### DIFF
--- a/cc-ansible
+++ b/cc-ansible
@@ -6,18 +6,18 @@ set -o pipefail
 # set -o xtrace
 
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
-VIRTUALENV="$(realpath "${VIRTUALENV:-"$DIR"/venv}")"
+VIRTUAL_ENV="$(realpath "${VIRTUAL_ENV:-"$DIR"/.venv}")"
 
 FORCE_UPDATES="${FORCE_UPDATES:-no}"
-if [[ ! -d "$VIRTUALENV" ]]; then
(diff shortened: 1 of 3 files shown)


cc-ansible
```diff
@@ -365,15 +356,7 @@ kolla_args+=(--globals="$CONFIG_DIR/globals.yml")
...
 kolla_args+=(--extra node_custom_config="$CONFIG_DIR/node_custom_config")
 kolla_args+=(--extra cc_ansible_site_dir="$CONFIG_DIR")
-# Override python interpreter to point to virtualenv (but only if we're
-# not performing the bootstrap, which is responsible for setting up the
-# virtualenv in the first place!)
-if [[ ! "${POSARGS[*]}" =~ bootstrap-servers ]]; then
-  # Run a quick ad-hoc command to figure out the derived value of the
-  # 'virtualenv' variable, as it can differ by site.
-  virtualenv=$(get_defined_var virtualenv)
-  kolla_args+=(--extra ansible_python_interpreter="$virtualenv/bin/python")
```
diurnalist, line 375: This point of this code is to ensure that Ansible modules _on the target host_ run inside _the virtualenv created by Kolla._ That's a different type of virtualenv than the rest of this patch addresses. Did you test this on an environment where VIRTUAL_ENV on the deploy node != VIRTUAL_ENV on the target?

PR #130, opened 2021-10-14
Fixup issues with chameleon-specific roles

diurnalist:
Many of the roles were missing "become: yes" for the sudo use-case.
The mariadb backup role was missing some defaults and also wasn't
creating the backup user.

For the heat download task, I made it so you can plop a google JSON
key (encrypted w/ vault) into the site config and it will use that.


kolla/node_custom_config/ironic.conf
```diff
@@ -3,7 +3,6 @@ enabled_console_interfaces = ipmitool-shellinabox,ipmitool-socat,no-console
 default_console_interface = ipmitool-socat
 enabled_network_interfaces = flat,neutron
 default_network_interface = neutron
-my_ip = "{{ kolla_internal_vip_address }}"
```
msherman64, line 6: It looks like this defaults to:
`my_ip = {{ api_interface_address }}` In the kolla-ansible template.

We might need to manually override this with something in defaults.yml, to allow manually specifying the iface.

msherman64 (approved):

PR #131, opened 2021-10-14
Refactor example site config to set up dev-friendly site

diurnalist:
I was playing around with deploying CHI-in-a-box on a VM for evaluation/dev purposes and landed on a config that seems to work ok and removes some of the requirements. I'm putting them here in the example site config, maybe there is a better place?

### DIFF
--- a/site-config.example/defaults.yml
+++ b/site-config.example/defaults.yml
@@ -1,30 +1,47 @@
 ---
-# This file contains an example site configuration.
-# To enable features, each section MUST be customized to you needs.
+# This file contains a minimum site configuration that brings up all services
+# on local connections for testing.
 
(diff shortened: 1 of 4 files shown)


msherman64 (commented):
This looks like a reasonable set of changes.

As for where to put it, perhaps in https://github.com/ChameleonCloud/site-config-template ?
We'd need to make sure that `cc-ansible init` understands how to do the needed find+replace though.

msherman64:
Good idea with the `.gitkeep` files, didn't know about those.

diurnalist:
@msherman64 `.gitkeep` is just a convention that folks have come up with, it's not something that Git really understands or cares about, other than being a file! But being hidden helps.

PR #132, opened 2021-10-14
Fix python interpereter for deploy-host usecase.

msherman64:
don't set python interpereter via CLI, since it applies to all nodes.
This fails when the deploy host has a different venv path than
the target hosts.
Default is set in kolla/defaults.yml anyway.

### DIFF
--- a/cc-ansible
+++ b/cc-ansible
@@ -186,14 +186,6 @@ USAGE
   exit 1
 }
 
-get_defined_var() {
-  local varname="$1"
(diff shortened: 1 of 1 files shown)


diurnalist (approved):
OK... I feel like I have memories of this code slamming in and out of this file or otherwise being removed and re-added a lot. We'll see :)

msherman64:
I'm not sure of the "right way" to do this, but when it's present, it tries to use `/etc/ansible/venv` on the deploy-host, and overrides any python path set in host/group vars.

PR #134, opened 2021-11-04
use kolla-ansible mariadb backups

zhenz:


roles/chameleon_mariadb/tasks/deployment.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Add mariadb backup job
+  include_role:
+    name: chameleon.periodic_task
+  vars:
+    task_name: "mariadb_backup_{{ mariadb_backup_timer_suffix }}"
+    task_command: "sh {{ deployment_dir }}/cc-ansible --site {{ deployment_site_config_dir }} --tags mariadb_backup"
```
diurnalist: I don't think the `sh` is necessary here

roles/chameleon_mariadb/tasks/site.yml
```diff
@@ -0,0 +1,37 @@
+---
```
diurnalist: I think a better name for this task is `config`, as it is configuring the backup script on the MariaDB host.

roles/chameleon_mariadb/templates/mariadb.j2
```diff
@@ -9,13 +9,17 @@
 # FLUSH PRIVILEGES;
 #
 
+ROTATE="{{ mariadb_backup_file_age }}"
+
+set -o pipefail
+
+{% if not enable_mariabackup %}
```
diurnalist, line 16: Do we need to keep this still? What if we removed the old way instead of preserving the two different ways of doing this?

roles/chameleon_mariadb/templates/mariadb.j2
```diff
@@ -32,21 +36,34 @@ ADDITIONAL_OPTIONS="--ignore-table=mysql.event"
...
+
+{% if enable_mariabackup %}
+
+DIR=$(docker volume inspect mariadb_backup | jq -r '.[].Mountpoint')
+PREFIX=mysqlbackup-
+backup_file="${DIR}/$(ls -t1 ${DIR}/${PREFIX}* |  head -n 1)"
+
+cleanup() {
+  # Rotate backups
+  find "${DIR}/" -maxdepth 1 -type f -name "${PREFIX}*" -mtime +${ROTATE} -print0 | xargs -0 -r rm -f
```
diurnalist: This may not work very well if the volume is not bind-mounted to the host (e.g., it could be provided by NFS or RBD or something). It may be better to run this through Docker:

```shell
docker run --rm -v mariadb_backup:/backup {{ kolla_toolbox_image_full }} \
  find /backup -maxdepth 1 -type f -name "${PREFIX}*" -mtime +${ROTATE} -print0 \
  | xargs -0 -r -rm -f
```

The above creates a new instance of a toolbox container (could be any base container, maybe even the mariadb container image, which tehre should be a variable for in scope) and mounts the backup at /backup, then deletes it inside the container.

roles/chameleon_mariadb/tasks/deployment.yml
```diff
@@ -0,0 +1,8 @@
...
+    task_calendar: "23:00"
```
diurnalist, line 8: `daily`?

playbooks/backups.yml
```diff
@@ -1,5 +1,14 @@
 ---
+- hosts: deployment
+  roles:
+    - role: chameleon_mariadb
+      when: hostvars[groups['mariadb'][0]]['enable_mariabackup'] | bool
```
diurnalist: Why not just `enable_mariabackup | bool`?

roles/chameleon_mariadb/defaults/main.yml
```diff
@@ -1,6 +1,9 @@
 ---
 backup_database_user: backup
 
+deployment_dir: hostvars[groups['mariadb'][0]]['deployment_dir']
```
diurnalist: I'm not sure why these variables are needed, couldn't it refer to just the value of `deployment_dir` and `site_config_dir`? I don't think the various hosts are setting different values for these.

msherman64 (approved):
This looks good to me, I don't see anything that Jason didn't comment on.
Thanks!

roles/chameleon_mariadb/templates/mariadb.j2
zhenz: I think TACC has to use the old way since its MariaDB is managed externally (i.e. not deployed via kolla), right?

roles/chameleon_mariadb/tasks/deployment.yml
zhenz: I set the "backup" script to run at 23:00, and the "copy to central location" script to 00:00. I'd like to specify the actual time here to make sure one runs before the other.

roles/chameleon_mariadb/templates/mariadb.j2
diurnalist: Grr, you're right, I thought they were on the containerized DB already.

diurnalist (approved):

PR #135, opened 2021-12-07
[blazar] add overwrite external service endpoints

zhenz:


kolla/node_custom_config/blazar.conf
```diff
@@ -7,7 +7,18 @@ exempt_projects = {% for project_id in blazar_usage_project_exemptions %}{{ proj
 
 {% if enable_blazar_allocation_enforcement | bool %}
 enabled_filters = MaxReservationLengthFilter, ExternalServiceFilter
+{% if blazar_external_service_endpoint is defined %}
```
diurnalist, line 10: I think we should also default this to some values in the defaults.yml, b/c most sites are not going to deviate.

```yaml
blazar_external_service_check_create_endpoint: "{{ chameleon_portal_url }}/balance/check-create/"
... etc
```

diurnalist (changes requested):

msherman64 (commented):
Looks good to me. Does this require any changes to the site config?

zhenz:
> Looks good to me. Does this require any changes to the site config?

I don't think so, as Jason asked me to add defaults.

msherman64 (approved):

diurnalist (approved):

PR #138, opened 2021-12-10
add v3 to auth url, since we need it?

msherman64:
v3 doesn't appear to be included in the auth_url variable
not having this will cause a 404 when connecting with the v3oidcpassword auth_type


diurnalist (approved):
Sure - odd that Horizon doesn't add it but OK!

PR #139, opened 2021-12-10
Remove legacy roles

diurnalist:
Inspired by our meeting today :)


roles/reference_api/tasks/main.yml
```diff
@@ -1,13 +0,0 @@
...
-    name: chameleon_referenceapi_1
-    ansible_connection: docker
-  changed_when: false
-
-- name: Update reference repository inside reference API container.
-  delegate_to: chameleon_referenceapi_1
-  command: |
-    cd /var/db/g5k-api/reference-repository \
-      && git fetch \
-      && git pull
```
msherman64, line 13: This part of portal-camino now, right?

diurnalist: Sort of, I am not sure portal-camino has the fetch+pull, but it manages the lifecycle of the reference API container. The fetch+pull should go there, not here, IMO.

msherman64: sounds good to me. If it breaks, we're motivated to fix it properly :)

msherman64 (approved):

PR #140, opened 2021-12-14
set ansible_python_interpreter during bootstrap

msherman64:
set the path during bootstrap to allow creation of virtualenv.

This is the "inverse" of the old behavior where we set it during "not bootstrap", but allows us to set a default in group vars, then override here.

### DIFF
--- a/cc-ansible
+++ b/cc-ansible
@@ -360,6 +360,16 @@ kolla_args+=(--passwords="$CONFIG_DIR/passwords.yml")
 kolla_args+=(--extra node_custom_config="$CONFIG_DIR/node_custom_config")
 kolla_args+=(--extra cc_ansible_site_dir="$CONFIG_DIR")
 
+# We set ansible ansible_python_interpreter to the virtualenv inside site-config, and override here during bootstrap.
+# Overriding at the CLI here applies to all hosts, including the deploy host.
+# Bootstrap-servers creates a virtualenv at $VIRTUALENV, on all hosts except the deploy host, requiring a special case.
+# The old behavior was to set it here if NOT bootstrapping, but that fails for the deploy-host case.
+
+if [[ "${POSARGS[*]}" =~ bootstrap-servers ]]; then
+  # Assume python3 can be found in $PATH, only used for creating the virtualenv.
+  kolla_args+=(--extra ansible_python_interpreter="python3")
+fi
+
 if [[ "${POSARGS[*]}" =~ post-deploy ]]; then
   cp -f "$DIR/playbooks/post_deploy.yml" "$ansible_path/chi_in_a_box_post_deploy.yml"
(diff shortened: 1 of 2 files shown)


cc-ansible
```diff
@@ -360,6 +360,16 @@ kolla_args+=(--passwords="$CONFIG_DIR/passwords.yml")
...
 kolla_args+=(--extra cc_ansible_site_dir="$CONFIG_DIR")
 
+# We set ansible ansible_python_interpreter to the virtualenv inside site-config, and override here during bootstrap.
+# Overriding at the CLI here applies to all hosts, including the deploy host.
+# Bootstrap-servers creates a virtualenv at $VIRTUALENV, on all hosts except the deploy host, requiring a special case.
+# The old behavior was to set it here if NOT bootstrapping, but that fails for the deploy-host case.
+
+if [[ "${POSARGS[*]}" =~ bootstrap-servers ]]; then
+  # Assume python3 can be found in $PATH, only used for creating the virtualenv.
+  kolla_args+=(--extra ansible_python_interpreter="python3")
```
diurnalist, line 370: This will mean that the virtualenv folder on the target hosts (aka everything but the deploy host) will effectively not be used, right? In which case, why have the virtualenv at all? I thought the main thing it was used for was the kolla_docker library module, which interacts w/ the Docker socket, which uses the Python docker library.

How can this work if the docker pip module is not installed on the host already?

msherman64: Note that there's no negation in this check now.
The virtualenv folder get set via a line in kolla's defaults.yml, or the site config (need to add to this PR).

The cli arg here overrides it only during bootstrap, and we already have a requirement that all hosts have python3 available, so this shouldn't trigger the same path issue as specifying the virtualenv here.

diurnalist: Gotcha, thanks, not seeing the additional defaults threw me

msherman64: Got the defaults in now, good point to call out. I was still fuzzy on the best way to set this, but we don't currently expose the ability to set this through cc-ansible. If needed, a site can override both of these in site-config.

kolla/defaults.yml
```diff
@@ -12,7 +12,11 @@ kolla_install_type: source
 # Remove this if pulling CentOS8 containers
 openstack_tag_suffix: ''
 
+# Set according to https://docs.openstack.org/kolla-ansible/latest/user/virtual-environments.html#target-hosts
+# virtualenv specifies the path to the kolla virtualenv, while ansible_python_interpreter must be set explicitly
+# since it cannot be templated. This is overridden during bootstrap by cc-ansible.
 virtualenv: /etc/ansible/venv
+ansible_python_interpreter: /etc/ansible/venv/bin/python
```
diurnalist, line 19: Good comment about templating not working, I forgot about that, but it makes sense.

diurnalist (approved):
:shipit:

PR #141, opened 2021-12-15
Rename MaxReservationLengthFilter to MaxLeaseDurationFilter

Mark-Powers:
This was renamed when pushed upstream, but the old name `MaxReservationLengthFilter` was still used in our fork.


diurnalist (approved):
Good catch to update it here as well.

PR #145, opened 2022-01-03
Fix/contributors

msherman64:
We should have contribution guidelines in chi-in-a-box. The impetus comes from qualifying for the open source gitbooks plan, however we don't currently have a good way for the community to contribute patches and bugfixes.

My hope was to end up with a template thats generally useful across our repos, as well as more specific info for chi-in-a-box and associate sites.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,2 @@
+# Please don't be mean
+really.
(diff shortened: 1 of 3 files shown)


diurnalist (approved):
LGTM! I updated the code of conduct to the [Contributor Convenant](https://www.contributor-covenant.org/), which is used by many other open source projects.

PR #148, opened 2022-01-12
Add default resource property options

Mark-Powers:
We may want to consider setting `retry_allocation_without_defaults = False` for network reservations, if we don't want users to get stitched networks when all of the other ones are reserved.

### DIFF
--- a/kolla/defaults.yml
+++ b/kolla/defaults.yml
@@ -60,7 +60,10 @@ blazar_database_user: blazar
 blazar_database_port: "{{ database_port }}"
 blazar_email_relay: "127.0.0.1"
 blazar_api_allocation_extras: user_name
+blazar_host_enable_default_resources: no
 blazar_host_default_resource_properties: '["=", "\$node_type", "compute_skylake"]'
+blazar_network_enable_default_resources: no
+blazar_network_retry_without_default_resources: no
 blazar_network_default_resource_properties: '["=", "\$stitch_provider", "none"]'
 
 # Cinder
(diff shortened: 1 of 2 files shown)


diurnalist (commented):
As you point out, I think it's a good idea to set `retry_allocation_without_defaults = False` for network reservations, let's do it! 👍🏻 

What happens if the site deployer sets this to false or an empty string? How do they disable it?

Mark-Powers:
@diurnalist Setting this to empty string should be fine, treated as empty resource properties. I added a bool to configure if this in included in the config, though we could also just recommend setting it to the empty string.

msherman64 (approved):
Looks good to me!

Could you add some short docs on the purpose, usage, configuration and so on to the docs? https://app.gitbook.com/o/Zkch89lL259lOhVcksT8/s/YIA6byel3J5Joam2z3kJ/reference/resource-reservation/default-resource-properties

Mark-Powers:
I've added docs: https://app.gitbook.com/o/Zkch89lL259lOhVcksT8/s/YIA6byel3J5Joam2z3kJ/reference/resource-reservation/default-resource-properties

I explain in the docs how this can be disabled. I think it depends on the definition of "disabled" however. If the desire is for the operator to not need to configure it, the can follow the docs. If the intention is to prevent a user from making an unconstrained request (e.g. they don't include resource properties), then we would want to advise setting it to an invalid query (e.g. [$id = -1]) and turning retry to false.

We need to set `$stitchprovider = none` on the remaining networks before deploying. This should be simple to script.

diurnalist (approved):
:shipit:

PR #160, opened 2022-02-01
enable floating IP reservations by default

msherman64:
allow reservation of floating IPs on network named "public" by default.


diurnalist (approved):
:shipit:

PR #162, opened 2022-02-28
Add missing bool filter in horizon config

Mark-Powers:


kolla/node_custom_config/horizon/custom_local_settings
```diff
@@ -76,12 +76,24 @@ OPENSTACK_BLAZAR_FLOATINGIP_RESERVATION = {
   'network_name_regex': '{{ blazar_floatingip_reservation_network_regex }}'
 }
 {% endif %}
+{% if enable_zun | bool %}
 OPENSTACK_BLAZAR_DEVICE_RESERVATION = {
-  'enabled': {{ enable_zun }},
+  'enabled': True,
```
diurnalist: Does this work? It's probably a bit cleaner and easier to extend.

```jinja
OPENSTACK_BLAZAR_DEVICE_RESERVATION = {
    'enabled': {% enable_zun | bool %},
}
```

diurnalist (approved):

kolla/node_custom_config/horizon/custom_local_settings
Mark-Powers: Can you check again? I had that initially but wasn't convinced it works, after more investigation I think your suggestion does work.

diurnalist (approved):
LGTM

PR #163, opened 2022-03-01
update chameleon usage reporting

zhenz:
- moved core usage services to cels gitlab
- removed jupyter usage reporting
- updated physical node usage reporting


roles/chameleon_usage/tasks/main.yml
```diff
@@ -1,2 +1,106 @@
...
+      host: "%"
+      password: "{{ chameleon_usage_mysql_password }}"
+      priv: "{{ all_privileges | join('/') }}"
+  run_once: True
+  # Delegate to control node, which we can safely assume is running at least
+  # one Kolla service, and therefore already has the kolla_toolbox container
+  # running on the host.
+  delegate_to: "{{ groups['control'][0] }}"
+  vars:
+    all_privileges:
```
diurnalist, line 46: Nice use of task vars 👍🏻

roles/chameleon_usage/tasks/main.yml
```diff
@@ -1,2 +1,106 @@
...
+      - 'nova.*:SELECT'
+
+- name: Initialize chameleon_usage database
+  shell: |
+    docker run --rm --net=host \
+    -v "{{ chameleon_usage_config_dir }}/my.cnf:/etc/mysql/my.cnf" \
+    {{ chameleon_usage_docker_image }} setup_database --mysql-conf /etc/mysql/my.cnf 2>&1
+
+- name: Configure chameleon usage data extract docker bash wrapper
+  template:
```
diurnalist, line 65: This probably also needs `become: yes` as it's writing to /usr/local/

roles/chameleon_usage/templates/chameleon-usage-extract-data.j2
```diff
@@ -3,9 +3,12 @@
 docker run --rm --net=host \
     -v "{{ backup_dir }}:{{ backup_dir }}" \
     -v "{{ chameleon_usage_config_dir }}/my.cnf:/etc/mysql/my.cnf" \
+    -v "{{ chameleon_usage_config_dir }}/site.yaml:/etc/chameleon_usage/site.yaml" \
     -v "/var/run/docker.sock:/var/run/docker.sock" \
     {{ chameleon_usage_docker_image }} \
-    {{ usage_extract_command }} --site {{chameleon_site_name}} --backup-location {{ backup_dir }} \
+    {{ usage_extract_command }} --site {{openstack_region_name}} \
```
diurnalist: Why not keep it `chameleon_site_name`? Only reason is, it may not be expected that changing the region name will change/break usage reporting. In practice we depend on region name in a lot of places so I think that's OK, but I'm thinking that maybe in the future we will want to have `chameleon_site_name` be more like `chameleon_site_id`? Or something. Like, we have `uc` as the ID and the display name is `CHI@UC`. cc @msherman64

roles/chameleon_usage/templates/chameleon-usage-sanity-checks.j2
```diff
@@ -4,4 +4,8 @@ docker run --rm --net=host \
     -v "{{ chameleon_usage_config_dir }}/slack.json:/etc/chameleon_usage/slack.json" \
     -v "{{ chameleon_usage_config_dir }}/my.cnf:/etc/mysql/my.cnf" \
     {{ chameleon_usage_docker_image }} \
-    node_usage_sanity_checks --mysql-conf /etc/mysql/my.cnf --slack /etc/chameleon_usage/slack.json 2>&1
+    node_usage_sanity_checks --mysql-conf /etc/mysql/my.cnf \
+    {% if slack_api_url is defined %}
```
diurnalist, line 8: good that it's optional now!

roles/chameleon_usage/templates/site.yaml.j2
```diff
@@ -0,0 +1,7 @@
+---
+username: "{{ chameleon_usage_site_user }}"
+password: "{{ chameleon_usage_site_password }}"
+
+maintenance_project_id: {{ maintenance_project_id }}
```
diurnalist: I missed where this is set - is it something the operator has to set in site-configs? It would be nice if this was automatic somehow (or, used a project name rather than ID, making it generic across sites.)

roles/chameleon_usage/templates/slack.json.j2
```diff
@@ -1,7 +1,4 @@
 {
     "webhook": "{{ slack_api_url }}",
-    "hostname_names": {
-        "m01-07.chameleon.tacc.utexas.edu": "CHI@TACC",
-        "fx2-01.uc.chameleoncloud.org": "CHI@UC"
-    }
+    "hostname_names": {}
```
diurnalist: What about removing it altogether? I'm assuming this isn't used anymore.

diurnalist (approved):
I think it's fine to remove the JupyterHub exporter, thanks for flagging. It doesn't have a ton of useful information in it and we need to move that infrastructure anyways. It should re-emerge in the nimbus playbook for JupyterHub, I think.

Looks good! A few comments/questions but overall 👍🏻

roles/chameleon_usage/templates/chameleon-usage-extract-data.j2
zhenz: The outage tags are using region names, so I thought it would be easier to just use region names for everything else instead of changing back and forth. But if we prefer the site name, I'll make changes to the usage reporting scripts accordingly.

diurnalist: IDK, @msherman64 what do you think? I don't have strong feelings I guess. Good point about outage tags using the region name currently @zhenz .

msherman64: Is the `chameleon_site_name` primarily used for the referenceapi? I feel like the region name is more user readable, and makes sense here

roles/chameleon_usage/templates/site.yaml.j2
msherman64: by convention, I believe we've been using `openstack` for the admin project, and `chameleon` for the devops project, but this doesn't extend well to associate sites.

I agree that some automation would be nice, both for creation of the projects (if local), and making it generic.
I'm hesitant to add too many settings that require specific configuration at every site.

msherman64 (approved):
Looks good to me, I do think we need to come back to the question of what users and projects are "special" at each site.

PR #164, opened 2022-03-01
Add blazar_host_url_format in horizon config

Mark-Powers:
This is to link the blazar calendar rows to the hardware page.

### DIFF
--- a/kolla/defaults.yml
+++ b/kolla/defaults.yml
@@ -68,6 +68,7 @@ blazar_host_retry_without_default_resources: yes
 blazar_network_default_resource_properties: '["=", "\$stitch_provider", "none"]'
 blazar_network_retry_without_default_resources: no
 blazar_floatingip_reservation_network_regex: "[pP]ublic"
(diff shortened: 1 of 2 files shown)


diurnalist (approved):
Where is the default defined? To avoid errors running the playbook, it should be defaulted to empty string or something. It may even be the case that this value is going to be the same for all sites? In which case, we can configure it in the kolla/defaults.yml and it will just work for all deployments.

Mark-Powers:
@diurnalist It is not defined by default, but the conditional statement checks that it is defined. Blazar dashboard checks if it is defined as well, and if it is undefined doesn't link to anything.

It is different per site since the site name is part of the URL, e.g.
```
https://chameleoncloud.org/hardware/node/sites/uc/...
https://chameleoncloud.org/hardware/node/sites/tacc/...
```

diurnalist:
@Mark-Powers we do have this available:

https://github.com/ChameleonCloud/chi-in-a-box/blob/master/site-config.example/defaults.yml#L8

Mark-Powers:
@diurnalist Updated to use `chameleon_site_name`. If a site wants this disabled, they can override the value to the empty string.

msherman64 (approved):
👍

PR #165, opened 2022-03-14
template key_file location if not manually set

msherman64:
Populate ML2 conf with ssh key information, if set.
Password is now optional if key is used.

### DIFF
--- a/kolla/node_custom_config/neutron/ml2_conf.ini
+++ b/kolla/node_custom_config/neutron/ml2_conf.ini
@@ -66,9 +66,14 @@ token = {{ switch.auth.token }}
 {% else %}
 ip = {{ switch.address }}
 username = {{ switch.auth.username }}
+{% if switch.auth.password is defined %}
 password = {{ switch.auth.password }}
+{% endif %}
 {% if switch.auth.secret is defined %}
 secret = {{ switch.auth.secret }}
+{% endif %}
+{% if (switch.auth.private_key is defined) and (switch.ngs_config.key_file is not defined) %}
+key_file = /etc/neutron/ssh/id_{{ switch.name }}
 {% endif %}
 {% endif %}
 {% for key, value in (switch.ngs_config | default({})).items() %}


kolla/node_custom_config/neutron/ml2_conf.ini
```diff
@@ -66,10 +66,15 @@ token = {{ switch.auth.token }}
 {% else %}
 ip = {{ switch.address }}
 username = {{ switch.auth.username }}
+{% if switch.auth.password is defined %}
 password = {{ switch.auth.password }}
+{% endif %}
 {% if switch.auth.secret is defined %}
 secret = {{ switch.auth.secret }}
 {% endif %}
+{% if (switch.auth.private_key is defined) and (switch.ngs_config.key_file is not defined) %}
```
diurnalist, line 75: It seems like password/key_file are mutually exclusive, what about

```jinja
{% if switch.auth.password is defined %}
password = {{ switch.auth.password }}
{% elif switch.ngs_config.key_file is defined %}
key_file = {{ switch.ngs_config.key_file }}
{% elif switch.ngs_config.private_key is defined %}
{# private key file is written out and mounted as part of container config #}
key_file = /etc/neutron/ssh/id_{{ switch.name }}
{% endif %}
```

diurnalist (approved):

kolla/node_custom_config/neutron/ml2_conf.ini
msherman64: they're not necessarily mutually exclusive, as netmiko will fall back to password if the key fails for whatever reason (not found, invalid, switch rejects it, etc)

msherman64: I think I'll merge as is, due to two factors:
1. the password can still be useful in combination, as above
2. the `key_file` gets passed in via the default ngs kwargs, so it needs to be handled separately to avoid a double template.

diurnalist: Yep, sorry, that's fine :)

PR #166, opened 2022-03-21
fixes for chameleon usage reporting and mariadb backup

zhenz:


kolla/defaults.yml
```diff
@@ -31,14 +31,15 @@ switch_configs: []
 neutron_networks: []
 
 enable_mariadb: yes
+enable_mariabackup: yes
```
msherman64: are there any defaults that need to be set if this is auto-enabled?

kolla/defaults.yml
```diff
@@ -31,14 +31,15 @@ switch_configs: []
...
 enable_chrony: yes
 enable_fluentd: yes
 enable_memcached: yes
 memcached_max_memory: 2048
 enable_rabbitmq: yes
 
 # Automatic backups
-backup_dir: /var/db/backup
+backup_dir: /var/lib/docker/volumes/mariadb_backup/_data
```
msherman64: Why is the host path to the volume being used here? This path depends on the docker volume driver that's in use, and generally shouldn't be used directly, instead favoring a docker mount of the volume.

roles/chameleon_usage/tasks/main.yml
```diff
@@ -69,7 +69,7 @@
 
 - name: Fill in up-to-date data to chameleon_usage
   become: yes
-  command: "/usr/local/sbin/chameleon-usage-extract-chi-data"
+  command: "/usr/local/sbin/chameleon-usage-extract-chi-data &"
```
msherman64: As this depends on the mysqlbackup information, should it be run inside a container with that volume mounted?

msherman64 (changes requested):
It seems that there's already a usage_reporting container referenced in earlier commits?

This looks reasonable, but I'd prefer to run scripts inside said container and use docker volume mounts, rather than explicitly referencing the volume path on the host.

kolla/defaults.yml
zhenz: I don't think so.

cc-ansible
```diff
@@ -359,6 +359,8 @@ kolla_args+=(--globals="$CONFIG_DIR/globals.yml")
 kolla_args+=(--passwords="$CONFIG_DIR/passwords.yml")
 kolla_args+=(--extra node_custom_config="$CONFIG_DIR/node_custom_config")
 kolla_args+=(--extra cc_ansible_site_dir="$CONFIG_DIR")
+kolla_args+=(--extra deployment_dir="$DIR")
+kolla_args+=(--extra site_config_dir="$CC_ANSIBLE_SITE")
```
msherman64, line 363: how are these two values used? What about the case where the user has set `--site </path/to/site_config>` rather than setting the environment variable `CC_ANSIBLE_SITE`?

zhenz: I think you can do either way:
- https://github.com/ChameleonCloud/chi-in-a-box/blob/master/cc-ansible#L192-L202
- https://github.com/ChameleonCloud/chi-in-a-box/blob/master/cc-ansible#L41

both `deployment_dir` and `site_config_dir` are used for setting up a periodic job of mariadb_backup: https://github.com/ChameleonCloud/chi-in-a-box/blob/master/roles/chameleon_mariadb/tasks/deployment.yml

msherman64: thanks, looks good to me!

msherman64 (approved):

PR #168, opened 2022-03-29
support for glance multi store

zhenz:


kolla/node_custom_config/glance.conf
```diff
@@ -1,5 +1,28 @@
+[DEFAULT]
+# TODO: remove define enabled_backends after upgrade kolla-ansible to xena
+{% if glance_store_backends %}
+enabled_backends = {% for key in glance_store_backends %}{{ key.name }}:{{ key.type }}{% if not loop.last %}, {% endif %}{% endfor %}
```
msherman64, line 4: should this be `glance_backends` instead of `glance_store_backends`?

kolla/node_custom_config/glance.conf
```diff
@@ -1,5 +1,28 @@
...
+{% endif %}
+
 [glance_store]
 os_region_name = {{ openstack_region_name }}
+default_backend = file
+
+[file]
+# TODO: remove file section after upgrade kolla-ansible to xena
+filesystem_store_datadir = /var/lib/glance
```
msherman64: this looks like it should be `filesystem_store_datadir = /var/lib/glance/images/`

which is already set in the kolla defaults https://github.com/ChameleonCloud/kolla-ansible/blob/dbe349d720eaef838597845e7d1202a163bf79d5/ansible/roles/glance/templates/glance-api.conf.j2#L57

It needs to store images in a subdirectory of the volume, since the volume is also used for image caching and other purposes.

kolla/node_custom_config/glance.conf
```diff
@@ -1,5 +1,28 @@
...
+[swift]
+swift_store_container = chameleon-images
+swift_store_multiple_containers_seed = 0
+swift_store_multi_tenant = False
+swift_store_create_container_on_put = True
+swift_store_region = {{ openstack_region_name }}
+default_swift_reference = swift
+swift_store_config_file = /etc/glance/glance-swift.conf
+swift_store_auth_insecure = True
+{% endif %}
```
msherman64, line 25: Is it intended that all sites have write access to the swift container? If the swift bucket is set to public, sites could access it read only via the http store type.

kolla/defaults.yml
```diff
@@ -83,6 +83,17 @@ enable_doni: yes
 
 # Glance
 enable_glance: yes
+glance_file_datadir_volume: /var/lib/glance
```
msherman64, line 86: Could you add this line, with a comment, to both the example configuration, and the documentation?
A couple of sites are interested in placing this directory on shared storage, and that's currently not in our documentation.

upstream docs: https://docs.openstack.org/kolla-ansible/train/reference/shared-services/glance-guide.html#file-backend

msherman64 (changes requested):
overall looks good, just want to doublecheck variable naming and add docs.

kolla/node_custom_config/glance.conf
zhenz: the kolla defaults has it under the `glance_store` section, but for multi-store, I need to put it under the `file` section. I've updated it to `/var/lib/glance/images/`.

kolla/node_custom_config/glance.conf
zhenz: the upstream has it as `glance_store_backends`

kolla/node_custom_config/glance.conf
zhenz: we'd like to have the `chameleon-images` swift container be public -- all sites can list the container and download from the container without credetials, but they can't upload to the container.

kolla/defaults.yml
zhenz: added.

msherman64 (commented):
Looks good to me, thanks!

msherman64 (approved):

PR #173, opened 2022-04-12
Update defaults.yml

QuantumTux:
This is a well-annotated version with explanatory text for each option. Where appropriate it references the GitBook documentation

### DIFF
--- a/site-config.example/defaults.yml
+++ b/site-config.example/defaults.yml
@@ -1,75 +1,103 @@
 ---
-# This file contains an example site configuration.
-# To enable features, each section MUST be customized to you needs.
+# /opt/chi-in-a-box/site-config/defaults.yml 
+# This file defines your basic site configuration
+# IMPORTANT: You MUST provide valid values for items marked "MANDATORY"
+# Some sections may be further customized to your needs
 
-# Associate Site Name (MANDATORY)
-openstack_region_name: CHI@XYZ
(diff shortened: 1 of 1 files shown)


site-config.example/defaults.yml
```diff
@@ -1,75 +1,103 @@
...
+# Otherwise, enable the appropriate federation service
+enable_keystone_federation: no
+enable_keystone_federation_openid: no
+keystone_idp_client_id: null
+###########################################################################
+
+###########################################################################
+# MANDATORY: Neutron networks
+# At minimum, you MUST define:
+#   0) A "public" network, corresponding to "eno1"
```
diurnalist, line 73: nit: it may not always be eno4, I'm wondering how best to communicate what this value should be. For the public network, it needs to be on an access interface (i.e., not trunked) that can reach the Internet w/o NAT. For the private network, it should be on an interface that can reach the top-of-rack switches, both for bare metal and for KVM configurations.

@msherman64 any suggestions?

diurnalist (commented):
Thanks a lot for the improvements! @msherman64 I know you had started moving some of the default/example config to another repo, does it make sense to integrate these changes there too?

msherman64:
I've integrated many of the suggestions here in a new PR https://github.com/ChameleonCloud/chi-in-a-box/pull/199
Closing this one. Thank you for the input!

PR #175, opened 2022-04-26
add doc for associate site tools and fixes

zhenz:


msherman64 (approved):
Looks good to me. Could you make the commit message more descriptive when merging? This has some changes to roles / defaults that aren't just docs.

PR #177, opened 2022-05-02
Add openstackclient tools to venv

msherman64:
Add openstackclient to the venv installation. A common question is how to install/run the tools, and this is sufficient given that the admin-openrc is present.
Installing climeleon seems overkill here.


diurnalist (approved):
Over time I think we'll be better served if "overkill" doesn't describe the process of installing and setting up climeleon, but this is a sane improvement! 🚢

PR #184, opened 2022-05-19
changes for filesystem

zhenz:
- add a filesystem role and include in post deploy
- filesystem role will create the default share type and ganesha
  external network
- add manila conf file and policy file
- add horizon custom local settings for manila

### DIFF
--- a/kolla/node_custom_config/manila.conf
+++ b/kolla/node_custom_config/manila.conf
@@ -1,0 +1,9 @@
+[DEFAULT]
+{% if default_share_type is defined %}
+default_share_type = {{ default_share_type }}
+{% endif %}
+quota_shares = 10
+quota_gigabytes = 2000
+
+[cephfsnfs1]
+cephfs_ganesha_export_ips =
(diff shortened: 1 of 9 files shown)


kolla/node_custom_config/manila.conf
```diff
@@ -0,0 +1,9 @@
+[DEFAULT]
+{% if default_share_type is defined %}
```
diurnalist, line 2: `is defined` will also be true if it's empty, I think? You can do `is defined and not empty` I think.

kolla/node_custom_config/manila/policy.json
```diff
@@ -0,0 +1,6 @@
+{
```
diurnalist, line 1: I think YAML is going to be the way going forward, can it be expressed as YAML?

roles/filesystem/tasks/main.yml
```diff
@@ -0,0 +1,96 @@
+---
+- name: Create default share type
+  become: true
+  command: >
+    docker exec kolla_toolbox openstack
```
diurnalist, line 5: It may be beneficial to use the `kolla_toolbox: ` module, which I think performs the same function but perhaps more reliably. You could run the openstack bit as a `command` module:

```
kolla_toolbox:
  module_name: command
  module_args: |
    openstack --os-interface ...
```

diurnalist (approved):
Great! ⭐

PR #185, opened 2022-05-23
fix admin-openrc to use site_config_dir

msherman64:
Rather than setting another arg via cc-ansible,set the path for the
site-config directory in the defaults.yml. This makes debugging easier.

Then, use this value to set the path for the admin-openrcr, to ensure
it is placed outside the tmp dir.

### DIFF
--- a/kolla/defaults.yml
+++ b/kolla/defaults.yml
@@ -13,7 +13,7 @@ kolla_install_type: source
 openstack_tag_suffix: ""
 
 # Write admin openRC to the site configuration
-admin_openrc_directory: "{{ cc_ansible_site_dir }}"
+admin_openrc_directory: "{{ site_config_dir }}"
 
 # Set according to https://docs.openstack.org/kolla-ansible/latest/user/virtual-environments.html#target-hosts
 # virtualenv specifies the path to the kolla virtualenv, while ansible_python_interpreter must be set explicitly


diurnalist (approved):
Nice, but why do we need to default to the env lookup if it's always passed in?

msherman64:
Right now in cc-ansible, we both set the ENV var, as well as using that env var to pass an `--extra` ansible_var into kolla.
I would like to default to setting ansible_vars in the defaults.yml, rather than via command-line argument, as it's easier to debug and inspect.

The end goal would be to have cc-ansible only specify the minimum needed cli args to wrap kolla-ansible, and have the rest be "native" ansible with proper defaults and error checking.

However, I can move that refactoring to a later PR, and just update the ansible_var used for now.

diurnalist:
@msherman64 a few nits:

> Right now in cc-ansible, we both set the ENV var, as well as using that env var to pass an --extra ansible_var into kolla.

We do not set the env var (note it is not exported). The _user_ can decide to set the env var. Otherwise that var defaults to the value of the `--site` flag. The point of the env var is that the user can set it to avoid having to type `--site ...` each time.

The only way Ansible is going to get the value of that variable reliably is via the extra_var at the moment; it's not guaranteed the env is set.

> The end goal would be to have cc-ansible only specify the minimum needed cli args to wrap kolla-ansible, and have the rest be "native" ansible with proper defaults and error checking.

We set the following:

* node_custom_config: this is needed to find where the custom config is; we have to pass it in b/c it's a tempdir (i.e., can't be set in defaults.yml). It's in a tempdir because it is a result of merging this dir in the site config w/ the dir in chi-in-a-box.
* cc_ansible_site_dir: not sure this is needed anymore and really doesn't have value b/c of the tempdir issue. It was supposed to solve the problem this PR solves w/ site_config_dir
* deployment_dir: also not sure what this is doing tbh
* site_config_dir: refers to the (non-tmpdir) site config dir 👍🏻 

I am not sure how we get around NOT setting `node_custom_config` and `site_config_dir`, and I think those are probably the only 2 extra vars we need to set at all.

msherman64:
I was a bit too imprecise. We're using 3 kinds of variables, env vars, bash vars, and ansible vars.
While we don't set the CC_ANSIBLE_SITE env var, we do assume that the bash variable is valid.

After reading the source for kolla-ansible, I agree that setting these basics via the `extra_var` functionality makes sense, I just want to avoid duplication.

PR #189, opened 2022-06-07
remove (hopefully) deprecated roles and files

msherman64:
None of these should be needed anymore, please comment if not!

(And I'll break these up into a separate commit per service removed)


diurnalist (commented):
I think JupyterHub may have diverged from the ansible-playbooks repo we have for core services. Otherwise I think these can, yes, all go :)

codyhammock (approved):
Looking good 👍🏻

zhenz (approved):
No red flag to the best of my knowledge 👍

PR #194, opened 2022-06-14
changes for filesystem

zhenz:
- add a filesystem role and include in post deploy
- filesystem role will create the default share type and ganesha
  external network
- add manila conf file and policy file
- add horizon custom local settings for manila

### DIFF
--- a/kolla/node_custom_config/manila.conf
+++ b/kolla/node_custom_config/manila.conf
@@ -1,0 +1,11 @@
+[DEFAULT]
+{% if default_share_type is defined %}
+default_share_type = {{ default_share_type }}
+{% endif %}
+quota_shares = 10
+quota_gigabytes = 2000
+
+[cephfsnfs1]
+{% if ganesha_vlan_interface_address is defined %}
+cephfs_ganesha_export_ips = {{ ganesha_vlan_interface_address }}
+{% endif %}
--- a/roles/filesystem/defaults/main.yml
+++ b/roles/filesystem/defaults/main.yml
@@ -1,0 +1,6 @@
+---
+filesystem_address_scope: filesystem-address-scope
+filesystem_subnet_pool: filesystem-subnet-pool
+filesystem_ganesha_network: filesystem-ganesha
+filesystem_ganesha_router: filesystem-ganesha-router
+enable_manila_nfs_ganesha_service: "{{ enable_manila_backend_cephfs_nfs | bool }}"
--- a/roles/filesystem/tasks/main.yml
+++ b/roles/filesystem/tasks/main.yml
@@ -1,0 +1,128 @@
+---
+- name: Create default share type
+  become: true
+  command: >
+    docker exec kolla_toolbox openstack
+    --os-interface internal
+    --os-auth-url {{ openstack_auth.auth_url }}
+    --os-identity-api-version 3
+    --os-project-domain-name {{ openstack_auth.user_domain_name }}
+    --os-project-name {{ keystone_admin_project }}
+    --os-username {{ keystone_admin_user }}
+    --os-password {{ keystone_admin_password }}
+    --os-user-domain-name {{ openstack_auth.user_domain_name }}
+    --os-region-name {{ openstack_region_name }}
+    share type create {{ default_share_type }} false
+    --extra-specs vendor_name=Ceph storage_protocol=NFS
+  register: manila_default_share_type
+  changed_when: manila_default_share_type is success
+  failed_when:
+    - manila_default_share_type.rc != 0
+    - manila_default_share_type.stderr.find('already') == -1
+  run_once: True
+  when:
+    - default_share_type is defined
+    
+- name: Create filesystem address scope
+  kolla_toolbox:
+    module_name: openstack.cloud.address_scope
+    module_args:
+      auth: "{{ openstack_auth }}"
+      state: present
+      name: "{{ filesystem_address_scope }}"
+      ip_version: 4
+      shared: true
+      project: "{{ keystone_admin_project }}"
+  run_once: True
+  become: True
+  when:
+    - enable_manila_nfs_ganesha_service | bool
+
+- name: Create filesystem subnet pool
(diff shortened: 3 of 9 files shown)


kolla/node_custom_config/manila.conf
```diff
@@ -0,0 +1,9 @@
+[DEFAULT]
+{% if default_share_type is defined %}
+default_share_type = {{ default_share_type }}
+{% endif %}
+quota_shares = 10
+quota_gigabytes = 2000
+
+[cephfsnfs1]
+cephfs_ganesha_export_ips = {{ ganesha_vlan_interface_address }}
```
msherman64, line 10: Does this have a default?

roles/filesystem/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+filesystem_address_scope: filesystem-address-scope
+filesystem_subnet_pool: filesystem-subnet-pool
+filesystem_ganesha_network: filesystem-ganesha
+filesystem_ganesha_router: filesystem-ganesha-router
+enable_manila_nfs_ganesha_service: "{{ enable_manila_backend_cephfs_nfs | bool }}"
```
msherman64, line 6: is a default set here? I'm not sure what the default is in kolla-ansible

roles/filesystem/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+    --os-auth-url {{ openstack_auth.auth_url }}
+    --os-identity-api-version 3
+    --os-project-domain-name {{ openstack_auth.user_domain_name }}
+    --os-project-name {{ keystone_admin_project }}
+    --os-username {{ keystone_admin_user }}
+    --os-password {{ keystone_admin_password }}
+    --os-user-domain-name {{ openstack_auth.user_domain_name }}
+    --os-region-name {{ openstack_region_name }}
+    share type create {{ default_share_type }} false
+    --extra-specs vendor_name=Ceph storage_protocol=NFS
```
msherman64, line 16: I see that openstack.cloud doesn't have a module for this. I think this is OK for now, but we should consider either making our own module, or contributing to upstream.

roles/filesystem/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+  register: manila_default_share_type
+  changed_when: manila_default_share_type is success
+  failed_when:
+    - manila_default_share_type.rc != 0
+    - manila_default_share_type.stderr.find('already') == -1
+  run_once: True
+  when:
+    - default_share_type is defined
+    
+- name: Create filesystem address scope
```
msherman64, line 26:
```suggestion
- name: Create filesystem address scope
  become: true
```

roles/filesystem/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+      state: present
+      name: "{{ filesystem_address_scope }}"
+      ip_version: 4
+      shared: true
+      project: "{{ keystone_admin_project }}"
+  run_once: True
+  when:
+    - enable_manila_nfs_ganesha_service | bool
+
+- name: Create filesystem subnet pool
```
msherman64, line 41:
```suggestion
- name: Create filesystem subnet pool
  become: true
```

roles/filesystem/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+      maximum_prefix_length: 32
+      minimum_prefix_length: 26
+      shared: true
+      project: "{{ keystone_admin_project }}"
+  run_once: True
+  when:
+    - enable_manila_nfs_ganesha_service | bool
+    - filesystem_subnet_pool_prefixes is defined   
+
+- name: Create filesystem-ganesha network
```
msherman64, line 62:
```suggestion
- name: Create filesystem-ganesha network
  become: true
```

roles/filesystem/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+      shared: false
+      port_security_enabled: true
+      project: "{{ keystone_admin_project }}"
+      state: present
+  run_once: True
+  when:
+    - enable_manila_nfs_ganesha_service | bool
+    - filesystem_ganesha_network_vlan is defined
+
+- name: Create filesystem-ganesha subnet
```
msherman64, line 83:
```suggestion
- name: Create filesystem-ganesha subnet
  become: true
```

roles/filesystem/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+      enable_dhcp: true
+      ip_version: 4
+      project: "{{ keystone_admin_project }}"
+      state: present
+  run_once: True
+  when:
+    - enable_manila_nfs_ganesha_service | bool
+    - filesystem_ganesha_network_vlan is defined
+
+- name: Create filesystem-ganesha router
```
msherman64, line 110:
```suggestion
- name: Create filesystem-ganesha router
  become: true
```

msherman64 (changes requested):
A few minor change, otherwise looks good!

1. We want to ensure that the role doesn't error out if defaults aren't set.
2. Invocations of docker exec or kolla_toolbox need `become: true`. The user executing cc-ansible may not be a member of the docker group.

roles/filesystem/defaults/main.yml
zhenz: kolla-ansible has default as false. https://github.com/ChameleonCloud/kolla-ansible/blob/chameleoncloud/xena/ansible/group_vars/all.yml#L692

roles/filesystem/tasks/main.yml
zhenz: I'll have it as it is for now. will keep in mind updating this in the future.

kolla/node_custom_config/manila.conf
zhenz: fixed with `if ganesha_vlan_interface_address is defined`

PR #195, opened 2022-06-23
clean up glance.conf

zhenz:


msherman64:
Are any changes needed to `kolla/defaults.yml` as well, or just this?

zhenz:
> Are any changes needed to `kolla/defaults.yml` as well, or just this?

@msherman64 TACC needs to have "glance_backend_swift: yes", but I don't think it should be in `kolla/defaults.yml`. only tacc needs to have multi-store enabled as "centralized image location".

msherman64 (approved):
Looks good to me, thanks!

PR #196, opened 2022-06-23
Fix vendordata post-deploy roles

msherman64:
This fixes two issues.
1. the syntax to become a specific user is no longer valid, but is
   not needed, as the role uses become where necessary
2. Ensure the docker image is downloaded. The controller node may not be
   logged into the chameleon docker registry, in which case the systemd
   unit will fail to download the iamge or start.


codyhammock (approved):
LGTM! 👍🏻

PR #197, opened 2022-06-23
ensure image tools can download container

msherman64:
Log into the registry to allow image pulling.


codyhammock (approved):
Looks straightforward

PR #199, opened 2022-06-24
document possible options in defaults.yml

msherman64:
This is an attempt to make the site-configuration more explicit, and demonstrate both a very minimal example, as well as an annotated case.

### DIFF
--- a/site-config.example/defaults.yml
+++ b/site-config.example/defaults.yml
@@ -11,63 +11,160 @@
 # And editing a host_vars file for host-specific overrides
 
 
-# Associate Site Name (MANDATORY)
-openstack_region_name: CHI@XYZ
-# Site name, similar to region but used for out-of-band inventory management
-chameleon_site_name: xyz
+#####################################
+# Base Configuration
+#####################################
 
-# Set to match the OS of your controller node(s)
+# Entries in this section must be set before running bootstrap-servers
+# The represent the minimal values needed to bring up the control-plane
+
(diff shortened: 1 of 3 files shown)


QuantumTux:
**VERY good update!** I have a few additional suggestions

Typo:  **The** represent the minimal values needed to bring up the control-plane
The first word should be "They"

If **glance_file_datadir_volume** does not ordinarily need to be changed from the default, then state that; describing what it is doesn't reveal if it is likely to need changing

Passive voice: Re-word "This interface is what all your api services will be bound to by default." to "This interface is where API services bind by default" 

Multiple places state "This interface must contain an IP address." - _contain_? I think it is clearer to say "An IP address must be bound to the interface." (if that is what is meant)

The text around **kolla_internal_vip_address** should probably make clear that yes, it must be defined even if there is only a single Control Node and the environment is not actually using HA

Reword "Entries below this point do not need bootstrap-servers to be rerun" - I suggest "After bootstrap-servers has been successfully executed, if you make changes to any configuration items below this point, you do not need to re-run bootstrap-servers" (although, what **DOES** someone need to do to make the changes effective?)

Under the "Associate Site" section, there is a line ending "...for Chameleon Associate sites, but can be ignored for independent use." Change "independent use" to "Independent Testbeds" so the term is used consistently.

msherman64:
@tuxrenci thank you for the comments! I've tried to address those in the latest commits.

PR #201, opened 2022-07-01
Document how to set default ironic images

msherman64:
allow setting default ironic images in defaults.yml
These will be used if a per-node value is not set.


codyhammock (approved):
LGTM

zhenz (approved):
👍

diurnalist (approved):

PR #206, opened 2022-09-15
deploy chi@edge v2

msherman64:
This PR allows for deployment of a Kubernetes control plane via K3s onto all control nodes.
Additional worker nodes can be added as per https://rancher.com/docs/k3s/latest/en/quick-start/
Zun will be configured to use said K3s cluster to deploy containers onto these worker nodes.

Note: You'll need to deploy in the following sequence:
1. Run a full deploy with `enable_k3s` set to false.
2. Set `enable_k3s` to true.
3. Run `cc-ansible --playbook ./playbooks/k3s.yaml`
4. Run `cc-ansible deploy` again. The second pass is needed because zun depends on having a token for the k3s cluster.

PR #216, opened 2022-09-24
host-endpoint: deny traffic to private ips by default

msherman64:
this globalnetwork  policy can be used to prevent egress from a
host to private IPs. This policy targets hostendpoints, which in turn
define which host interfaces are affected.

As it affects flows passing through said host interfaces, it acts to
restrict policy that targets workloads, rather than interfaces.

Example:

Namespaced networkpolicy permits egress traffic from container1 to container2
within the same namespace. As the destination is a known endpoint to calico,
it is included in "selector: all()" in the allow block of this policy, and the
Deny rule is not applied, despite traffic going to a private IP.

However, traffic from said container1 OR from services on the host itself to a
private IP not present in kubernetes, will be blocked by the second rule.

WARNING:
It's not clear what will happen in the case where a kubernetes endpoint exists that
overlaps with a local IP address.

### DIFF
--- a/docs/reference/resource-reservation/README.md
+++ b/docs/reference/resource-reservation/README.md
@@ -27,13 +27,17 @@ openstack reservation host create
 
 ### Devices
 
+```
+openstack reservation device create
+```
+
 ### Networks
 
 #### Create a Reservable Network
 
 This creates an entry in Blazar, allowing a user to reserve a vlan on a physical network. When their lease starts, a neutron network will be created using that vlan and physnet.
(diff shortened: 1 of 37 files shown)


msherman64:
Rebased onto https://github.com/ChameleonCloud/chi-in-a-box/pull/206 for a cleaner PR. Needs to include documentation on new defaults for network policy, and how users / operators can interact with it.

msherman64:
closing as duplicate

PR #229, opened 2022-09-29
Update minimal configuration example with dev mode

### LINKED ISSUE #228: Add defaults.yml for minimal development configuration
Getting set up for development can be time consuming. Add a default configuration that requires minimal tweaking to get started.

msherman64:
This updates the minimal example to use a loopback API interface, for
consistency across hosts. It also documents the procedure to enable
dev mode for all or specific services.


site-config.example/dev.defaults.yml
```diff
@@ -0,0 +1,124 @@
...
+# The values represent the minimal values needed to bring up the control-plane and API interfaces
+
+# All services will bind to this interface, on the INTERFACE IP address
+# This example assumes that `lo` has the IP `127.0.0.1/8`
+network_interface: "lo"
+
+# Keepalived will manage this IP address, which must be in the same
+# subnet as the interface IP above. Haproxy will bind to this IP, and the
+# openstack ADMIN and INTERNAL APIs listen here.
+kolla_internal_vip_address: "127.0.0.254"
```
msherman64, line 22: Turns out this breaks keepalived!

Due to the keepalived config: https://github.com/ChameleonCloud/kolla-ansible/blob/19eb03834a543a013ed820a7b370ec99e2825616/ansible/roles/loadbalancer/templates/keepalived/keepalived.conf.j2#L11

It will attempt to use the loopback as the primary vrrp interface, and fail to start. This config is fine for the case where haproxy is disabled, but not otherwise.

A workaround is to create a vlan subinterface, e.g. eth3.1000 to use instead.
The only requirment is that the interface is somehow "valid" for keepalived to watch.

PR #233, opened 2022-10-05
add sso config for grafana

msherman64:
This PR adds single-sign-on for Grafana.

TODO: Template list of allowed groups for admin, editor, and viewer roles

NOTE: we cannot allow public  access to the primary grafana instance/org, as viewers can run any possible query on any datasource. See https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/#limit-viewer-query-permissions

### DIFF
--- a/kolla/node_custom_config/grafana.ini
+++ b/kolla/node_custom_config/grafana.ini
@@ -1,0 +1,21 @@
+[server]
+root_url = "https://{{kolla_external_fqdn}}:3000"
+
+[auth]
+disable_login_form = "true"
+oauth_auto_login = "true"
+
+[auth.generic_oauth]
+name = Chameleon
+icon = signin
+enabled = true
+allow_sign_up = true
+client_id = "{{ keystone_idp_client_id }}"
+scopes = openid profile email
+auth_url = "https://auth.chameleoncloud.org/auth/realms/chameleon/protocol/openid-connect/auth"
+token_url = "https://auth.chameleoncloud.org/auth/realms/chameleon/protocol/openid-connect/token"
+api_url = "https://auth.chameleoncloud.org/auth/realms/chameleon/protocol/openid-connect/userinfo"
+
+role_attribute_path = contains(projects[*].id, 'Chameleon') && 'Admin' || contains(projects[*].id, 'CHI-210828') && 'Editor'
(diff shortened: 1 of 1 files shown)


codyhammock (approved):
LGTM

kolla/node_custom_config/grafana.ini
```diff
@@ -0,0 +1,21 @@
...
+icon = signin
+enabled = true
+allow_sign_up = true
+client_id = "{{ keystone_idp_client_id }}"
+scopes = openid profile email
+auth_url = "https://auth.chameleoncloud.org/auth/realms/chameleon/protocol/openid-connect/auth"
+token_url = "https://auth.chameleoncloud.org/auth/realms/chameleon/protocol/openid-connect/token"
+api_url = "https://auth.chameleoncloud.org/auth/realms/chameleon/protocol/openid-connect/userinfo"
+
+role_attribute_path = contains(projects[*].id, 'Chameleon') && 'Admin' || contains(projects[*].id, 'CHI-210828') && 'Editor'
```
Mark-Powers, line 19: what is the significance of 210828 here? Is this a project for associate site operators?

Mark-Powers (approved):
Looks good to me!

kolla/node_custom_config/grafana.ini
Mark-Powers: Actually, can this be templated? It seems strange to have this hardcoded.

msherman64, line 19: This line needs work.

mainly we need to be able to template a list of project IDs,
```
- Chameleon
- CHI-210828
- CHI-whatever
```
into some jmespath mapping onto roles. The easy answer is to skip the list idea and just have a "{{ site_operator_project }}" field that can be configured. Except, some of the operators don't have a longrunning project!

See https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/generic-oauth/#jmespath-examples

msherman64: exactly, see the note below. 210828 is from indyscc. I think the way to go would be to have a project for associate site operators, or perhaps one per site, similar to `Chameleon`. Perhaps "ops-<site_name>"

Issue #116, opened 2021-09-13
letsencrypt support requires manual action


msherman64:
Letsencrypt support currently requires manually requesting the first certificate.

This is because enabling letsencrypt support tells haproxy to use the TLS certificate, but letsencrypt runs after haproxy, and uses it for public access.

Therefore, enabling it only works if you already have a valid cert configured!

Currently, you can run the process manually, in this order.
1. Run deploy with letsencrypt disabled
2. use docker exec -it letsencrypt_certbot bash to manually run certbot
3. copy certificate out to site-config/certificates folder
4. enable letsencrypt, and re-run deploy.

super-cooper:
![](https://github.trello.services/images/mini-trello-icon.png) [Automate TLS certificate renewal for CHI-in-a-box sites](https://trello.com/c/nyJl1B03/2795-automate-tls-certificate-renewal-for-chi-in-a-box-sites)

msherman64:
Since we backported upstream's letsencypt integration, this is now resolved.
https://github.com/ChameleonCloud/chi-in-a-box/commit/925fc4737ca8b833da9f5a9b05de9ea631ab3dd8

PR #237, opened 2022-10-06
Automatically set `kolla_externally_managed_cert` when Let's Encrypt is enabled

super-cooper:
In general, users will want this enabled if they're using Let's Encrypt. It could cause confusion for new deployments, but this should no longer be an issue when #116 is figured out.


super-cooper:
[Automate TLS certificate renewal for CHI-in-a-box sites](https://trello.com/c/nyJl1B03/2795-automate-tls-certificate-renewal-for-chi-in-a-box-sites)

codyhammock (approved):

PR #238, opened 2022-10-07
add prometheus ironic exporter

msherman64:
This PR adds the [prometheus ironic exporter](https://opendev.org/openstack/ironic-prometheus-exporter) to export metrics about all nodes known to ironic.

This depends on the following PRs:
- [x] https://github.com/ChameleonCloud/kolla/pull/8
- [x] https://github.com/ChameleonCloud/kolla-containers/pull/32
- [x] https://github.com/ChameleonCloud/kolla-ansible/pull/31

### DIFF
--- a/kolla/defaults.yml
+++ b/kolla/defaults.yml
@@ -167,6 +167,10 @@ ironic_console_serial_speed: 115200n8
 encoded_ironic_pxe_root_password: "{{ ironic_pxe_root_password | password_hash('md5') |  regex_replace( '(\\$)', '$\\1') }}"
 ironic_pxe_append_params: nofb nomodeset vga=normal console=tty0 console=ttyS0,{{ ironic_console_serial_speed }} systemd.journald.forward_to_console=yes rootpwd="{{ encoded_ironic_pxe_root_password }}"
 
+# This can be "expensive"; allow setting it explicitly, but default to the prometheus interval
+ironic_send_sensor_data_interval: 60
+ironic_send_sensor_data_for_undeployed_nodes: False
+
 # number of times to retry failed neutron api requests
 # will use exponential backoff from 0.5s up to 60s, or
 # will use ironic_neutron_status_code_retry_delay if set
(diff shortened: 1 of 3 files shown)


msherman64:
This lets us correlate metrics with node id/name, instance id/name, user id/name, and project id/name

Example:
image

msherman64:
@diurnalist could I ask for a "big picture" review? It wasn't clear to me which changes should live in e.g. kolla vs kolla-contianers, or kolla-ansible vs chi-in-a-box. I tried to err on the side of:
* "basic functionality goes in kolla, kolla-ansible"
* "customization goes in kolla-containers, chi-in-a-box"

Mark-Powers (approved):
This looks good to me, and I agree with your reasoning for organizing these changes. I did not see anything of concern really.

kolla/defaults.yml
```diff
@@ -279,6 +283,9 @@ prometheus_server_external_url: "{{ public_protocol }}://{{ prometheus_external_
 prometheus_alertmanager_external_url: "{{ public_protocol }}://{{ prometheus_external_fqdn }}:{{ prometheus_alertmanager_port }}"
 prometheus_bind_address: "{{ lookup('vars', 'ansible_' + network_interface).ipv4.address }}"
 
+enable_prometheus_ironic_exporter: "{{ enable_prometheus | bool and enable_ironic | bool }}"
+# prometheus_ironic_exporter_interval: 60s
```
diurnalist: Maybe remove this commented line

diurnalist (approved):
LGTM!

diurnalist:
@msherman64 I think the way you organized it is spot on :) - I left a few little comments but yes, in general that's how I would approach such a patch. Try to put as much in K-A/Kolla as possible, treat it as if it were an upstream contribution, then add Chameleon-specific defaults in chi-in-a-box. Kolla-containers really only needs touching if you need to configure an additional container image to be built (as in this case.) In general I tried to keep overrides in kolla-containers really minimal, and it's probably the hackiest part of the whole thing :)

codyhammock (approved):
👍🏻

PR #239, opened 2022-10-10
allow mariadb backup defining user and group

zhenz:


codyhammock (approved):
👍🏻

roles/chameleon_mariadb/tasks/deployment.yml
```diff
@@ -12,3 +12,5 @@
     task_name: "mariadb_backup_{{ mariadb_backup_timer_suffix }}"
     task_command: "{{ deployment_dir }}/cc-ansible --site {{ deployment_site_config_dir }} mariadb_backup"
     task_calendar: "23:00"
+    task_user: "{{ mariadb_backup_task_user|default(None) }}"
+    task_group: "{{ mariadb_backup_task_group|default(None) }}"
```
msherman64, line 16: What happens if the task is run with "none" for these?

I couldn't find where the task_user and task_group are consumed.

msherman64: Ah, it's consumed in the other PR?
https://github.com/ChameleonCloud/ansible-role-periodic-task/pull/2

Appears to be "safe" in all cases, LGTM

msherman64 (approved):

PR #240, opened 2022-10-12
Add networkd-dispatcher hook for corsa NAT

msherman64:
Enable corsa NAT via networkd-dispatche hook.
This is needed to allow the corsa switches to "reach out" to openflow controllers running on public networks.

Required config variables:
- `corsa_nat_network_cidr`
- `corsa_nat_external_interface`

Actually creating the interface that corsas connect to (vlan3289 for UC), is out of scope for this PR, and should be done in e.g. the netplan or other persistent networking configuration.


mcevik0 (commented):
I do not have any useful knowledge about chi-in-a-box configurations. If a review from RENCI team is necessary, then @tuxrenci can be a better person for this since he already has some experience with chi-in-a-box. I will not be able to review this change.

PR #249, opened 2023-01-25
Zun: Workaround eventlet deadlock

msherman64:
When this is set to true (default as of Wallaby), a pthread is used for this task. # This currently (Xena) has a bad interaction with eventlet, causing occasional deadlocks. # This appears related to, but separate from this bug in oslo.log # https://bugs.launchpad.net/oslo.log/+bug/1983863


codyhammock (approved):
This seems simple enough. LGTM 👍🏻

PR #250, opened 2023-01-25
Add setting to discern baremetal sites

super-cooper:
This change introduces a new configuration parameter which allows sites to distinguish themselves as baremetal. This allows for distinct configuration between baremetal and other sites.

### DIFF
--- a/kolla/defaults.yml
+++ b/kolla/defaults.yml
@@ -28,6 +28,8 @@ chameleon_site_name: localhost
 # Whether to show a dropdown in the Horizon GUI that provides links to other
 # Chameleon testbed sites.
 enable_chameleon_multisite: yes
+# Is this site a baremetal site?
+chameleon_is_baremetal_site: no
 # Default networking fallbacks
 switch_configs: []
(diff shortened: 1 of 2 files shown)


super-cooper:
[Horizon - Remove unused tabs](https://trello.com/c/yRi5JHws/2825-horizon-remove-unused-tabs)

msherman64:
In some sense this is already defined by whether ironic is enabled:
https://github.com/ChameleonCloud/chi-in-a-box/blob/c12d27e0711735519371230a373bb4370bcb3a0c/kolla/defaults.yml#L160

Hypothetically in the future, we would like to allow 'hybrid' sites, that enable both KVM and baremetal, so making this granular would be handy.

Mark-Powers (approved):
I think the way this is implemented makes sense, as if there is a hybrid site in the future, with ironic enabled, this can still be toggled off. And we still support the current use case of toggling these for KVM vs fully baremetal sites.

codyhammock (approved):
No objections from me.

PR #255, opened 2023-03-07
allow cc-ansible to use local checkout of kolla-ansible

msherman64:
Adds environment variable KOLLA_ANSIBLE_LOCAL
If set, chi-in-a-box will attempt to install kolla-ansible in "editable" mode from this path, rather than from a remote git ref

useful for local development without a git push + pull


codyhammock (approved):
LGTM

super-cooper (approved):

PR #256, opened 2023-03-28
Fix ansible roles and keystone permissions for OSG

msherman64:
Our roles to deploy and enable OSG needed some fixing up.

The main changes are to properly create a keystone service user, and refer to the correct project domain when using said user.

Minor changes are to ensure that config and executables are owned by the correct users, and the roles can run as non-root (using `become` where necessary)


codyhammock (approved):
👍🏻

msherman64:
Closing for now as we dropped support/maintenance for OSG. Will revisit when we have the effort to support this feature again.

PR #263, opened 2023-08-29
enable ironic inspector

msherman64:
for ironic-inspector to work with neutron, the following need to be set:

enabled_inspect_interfaces must contain `inspector`
each node must have its inspect interface set to `inspector` by
running `openstack baremetal node set --reset-inspect-interface --inspect-interface inspector`

OR, if `default_inspect_interface` is set to `inspector`, new nodes will have their interface
set to this automatically, as well as when `--reset-inspect-interface` is called.

Finally, inspection_network MUST be set, or ironic will return a cryptic error message about how
"driver IPMI does not support managed inspection".

### DIFF
--- a/kolla/defaults.yml
+++ b/kolla/defaults.yml
@@ -164,10 +164,17 @@ ironic_provisioning_network: ironic-provisioning
 ironic_provisioning_network_cidr: 10.51.0.0/24
 ironic_provisioning_network_gateway: "{{ ironic_provisioning_network_cidr | ipaddr('next_usable') }}"
 ironic_cleaning_network: "{{ ironic_provisioning_network }}"
+ironic_inspection_network: "{{ ironic_provisioning_network }}"
 ironic_image_cache_size: 20480 # MB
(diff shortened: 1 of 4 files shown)


msherman64:
Before this can be merged, we need to be a little careful about what goes in node_custom_config.

Ironic-inspector supports a "standalone mode" that runs without neutron or glance, so it complains if pxe images aren't present in its own directory. I'd rather not stick 500mb of binary into the site config if we can avoid it, so this is still a work in progress ;)

codyhammock:
Can you be more specific about what settings in node_custom_config might be problematic?

msherman64:
> Can you be more specific about what settings in node_custom_config might be problematic?

Fixed the issues in this commit to kolla-ansible https://github.com/ChameleonCloud/kolla-ansible/commit/6917bd65e674791f4deeacea8e85cb1b453e6b44

Basically in the "standalone mode", ironic-inspector wants the ironic agent kernel+ramdisk in node_custom_config, but I was able to add a flag to skip this

PR #270, opened 2023-11-06
Upgraded the version of the Nvidia device plugin

JOUNAIDSoufiane:
Upgraded the NVIDIA Device plugin image to version 0.14.1 enabling support for running on L4T 32.7.2.

### DIFF
--- a/roles/k3s/files/nvidia-device-plugin.yaml
+++ b/roles/k3s/files/nvidia-device-plugin.yaml
@@ -48,9 +48,15 @@ spec:
       # See https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/
       priorityClassName: "system-node-critical"
       containers:
-        - image: nvcr.io/nvidia/k8s-device-plugin:v0.12.2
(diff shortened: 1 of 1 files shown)


roles/k3s/files/nvidia-device-plugin.yaml
```diff
@@ -48,9 +48,15 @@ spec:
...
+        - image: nvcr.io/nvidia/k8s-device-plugin:v0.14.1
           name: nvidia-device-plugin-ctr
-          args: ["--fail-on-init-error=false"]
+          env:
+            - name: FAIL_ON_INIT_ERROR
+              value: "false"
+            - name: NVIDIA_VISIBLE_DEVICES
+              value: all
+            - name: NVIDIA_DRIVER_CAPABILITIES
+              value: all
```
msherman64, line 59: Does the device plugin container need these flags set? I thought only the end-user containers needed it?

JOUNAIDSoufiane: yes it needs all these flags set as well, to see and access the GPU and to load the necessary Nvidia modules.  The fail on init error used to previously be a flag but is now set via an env var

msherman64 (approved):

PR #271, opened 2023-11-07
Add polling monitor flag for blazar Floating ip and new dry run flag for polling monitor

AnishReddyRavula:
- This is to enable the floating ip polling to clean up the resources in an errored or deleted lease
- * Setting the flag `enable_polling_monitor` to False then blazar does not periodically invoke the resource (FIP) monitor plugin to heal/clean the failed resources - PR for [FIP clean up](https://github.com/ChameleonCloud/blazar/pull/93)
- * When this option is not set, the monitor fails with following error - `oslo_config.cfg.NoSuchOptError: no such option enable_notification_monitor in group [virtual:floatingip]`

- Add config opt to blazar monitor to enable dry run
-  * Setting the flag `dry_polling_monitor` to True is to warn a failed/needs-fixing resource without attempting to fix it, to make sure that the fix is reliable and the monitor is actually picking up the right resource that needs fixing and the right fix is chosen

### DIFF
--- a/kolla/defaults.yml
+++ b/kolla/defaults.yml
@@ -76,6 +76,10 @@ blazar_network_default_resource_properties: '["=", "\$stitch_provider", "none"]'
 blazar_network_retry_without_default_resources: no
 blazar_floatingip_reservation_network_regex: "[pP]ublic"
 blazar_host_url_format: "https://chameleoncloud.org/hardware/node/sites/{{ chameleon_site_name }}/clusters/chameleon/nodes/{hypervisor_hostname}/"
+blazar_physical_polling_monitor: true
+blazar_physical_polling_monitor_dry_run: true
+blazar_fip_polling_monitor: true
+blazar_fip_polling_monitor_dry_run: true
 
 # Cinder
 enable_cinder: no
--- a/kolla/defaults.yml
+++ b/kolla/defaults.yml
@@ -341,6 +341,7 @@ blazar_enable_device_plugin_k8s: "{{ enable_k3s | bool }}"
 enable_etcd: "{{ enable_zun | bool and not enable_k3s | bool }}"
 enable_kuryr: "{{ enable_zun | bool and not enable_k3s | bool }}"
 enable_zun_compute: "{{ enable_zun | bool and not enable_k3s | bool }}"
+blazar_zun_polling_monitor: true
 
 # OSG
 enable_osg: no
--- a/kolla/node_custom_config/blazar.conf
+++ b/kolla/node_custom_config/blazar.conf
@@ -47,6 +47,7 @@ heartbeat_in_pthread = false
 before_end = email
 email_relay = {{ blazar_email_relay }}
 enable_polling_monitor = true
+dry_polling_monitor = true
 retry_allocation_without_defaults = {{ blazar_host_retry_without_default_resources | bool }}
 default_resource_properties = {{ blazar_host_default_resource_properties }}
 {% endif %}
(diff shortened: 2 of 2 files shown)


codyhammock (approved):
👍🏻

kolla/node_custom_config/blazar.conf
```diff
@@ -58,6 +58,7 @@ enable_polling_monitor = true
 
 [virtual:floatingip]
 billrate = {{ blazar_floatingip_billrate }}
+enable_polling_monitor = true
```
msherman64: What is the default behavior if this is not set?
Also, can you link to the blazar PR that depends on this flag?

AnishReddyRavula: If `enable_polling_monitor` is False for FIPs, then it won't use the FIP monitor plugin to heal resources ( in this case FIPs)

msherman64: Right, but what if it's not set at all?
Just want to be sure that blazar will start successfully if an older config is in use.

AnishReddyRavula: The monitor is set to run periodically and whenever it tries to run monitor, it fails with the following error
```oslo_config.cfg.NoSuchOptError: no such option enable_notification_monitor in group [virtual:floatingip]```

File: kolla/node_custom_config/blazar.conf [hunk 2]
```diff
@@ -47,6 +47,7 @@ heartbeat_in_pthread = false
 before_end = email
 email_relay = {{ blazar_email_relay }}
 enable_polling_monitor = true
+dry_polling_monitor = true
```
msherman64: This means that dry-run is enabled by default?
(optional) Maybe we can change the name of the flag to `enable_polling_monitor_dry_run`?

I would request that we set this via a variable, so that it can be set from defaults.yml, otherwise a site will need to use their own node_custom_config to override.

Something like:
```
enable_polling_monitor = {{ blazar_physical_polling_monitor }}
dry_polling_monitor = {{ blazar_physical_polling_monitor_dry_run }}
```

kolla/node_custom_config/blazar.conf
```diff
@@ -58,6 +59,8 @@ enable_polling_monitor = true
 
 [virtual:floatingip]
 billrate = {{ blazar_floatingip_billrate }}
+enable_polling_monitor = true
```
msherman64: We should add a variable for this also, e.g. 
```
enable_polling_monitor = {{ blazar_fip_polling_monitor }}
dry_polling_monitor = {{ blazar_fip_polling_monitor_dry_run }}
```

msherman64 (changes requested):

kolla/defaults.yml
```diff
@@ -76,6 +76,10 @@ blazar_network_default_resource_properties: '["=", "\$stitch_provider", "none"]'
 blazar_network_retry_without_default_resources: no
 blazar_floatingip_reservation_network_regex: "[pP]ublic"
 blazar_host_url_format: "https://chameleoncloud.org/hardware/node/sites/{{ chameleon_site_name }}/clusters/chameleon/nodes/{hypervisor_hostname}/"
+blazar_physical_polling_monitor: true
+blazar_physical_polling_monitor_dry_run: true
+blazar_fip_polling_monitor: true
+blazar_fip_polling_monitor_dry_run: true
```
msherman64, line 82: just add `blazar_zun_polling_monitor: true` and I think this is good.

msherman64 (approved):

PR #272, opened 2023-12-14
This is to enable the network polling to clean up the resources

AnishReddyRavula:
Setting the flag enable_polling_monitor to False then blazar does not periodically invoke the network monitor plugin to heal/clean the failed resources

When this option is not set, the monitor fails with following error - oslo_config.cfg.NoSuchOptError: no such option enable_notification_monitor in group

Add config opt to blazar monitor to enable dry run

Setting the flag dry_polling_monitor to True is to warn a failed/needs-fixing resource without attempting to fix it, to make sure that the fix is reliable and the monitor is actually picking up the right resource that needs fixing and the right fix is chosen

### DIFF
--- a/kolla/defaults.yml
+++ b/kolla/defaults.yml
@@ -80,6 +80,8 @@ blazar_physical_polling_monitor: true
 blazar_physical_polling_monitor_dry_run: true
 blazar_fip_polling_monitor: true
 blazar_fip_polling_monitor_dry_run: true
+blazar_network_polling_monitor: true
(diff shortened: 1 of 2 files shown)


msherman64 (approved):
LGTM. We just need to add docs for the new flags.

e.g.
" to enable the polling monitor for resource X, set blazar_network_polling_monitor_x (default true)"
"in dry run mode(default), the monitor will log warnings, but not make changes. to enable the polling monitor to correct issues found, set blazar_network_polling_monitor_dry_run: false"

PR #274, opened 2024-01-25
enable config_drive by default

msherman64:
enable the config_drive data source by default.

When enabled, this inject cloud-init metadata into an extra 64mb partition at the end of the boot disk.
Cloud-init (>v0.7.5) can automatically detect this config partition, and use it for pre-networking configuration when the network metadata endpoint is not yet available.

Specifically, this enables neutron network configuration to be dynamically sent on instance boot, letting users configure multiple interfaces, static ips, and more, rather than relying on DHCP on all interfaces.


msherman64:
edit: fixing up the change, it's supposed to be in the nova config file, not ironic

PR #275, opened 2024-01-27
mounted kubeconfig.yml

JOUNAIDSoufiane:


msherman64 (changes requested):
A couple requests:

1. please add a commit message describing what this is for, and link to the commit/pr in the hammers repo too
2. where is `hammers_kubeconfig_path` set, and what happens if it is not set?
3. if the file at `hammers_kubeconfig_path` is not present, what happens?

We want to make sure to "fail safe", as bind-mounting a non-existent path in docker has some odd behavior, in particular it will automatically create the path as a directory, and then future attempts to access it as a file will fail!

JOUNAIDSoufiane:
2. The hammers_kubeconfig_path is set under this [commit](https://github.com/ChameleonCloud/chi-in-a-box/blob/4e6f6e67cdcdf610321dc05d4277e4659ac3bdb9/roles/hammers/defaults/main.yml#L5)
3. This is the ansible error message if the file is not there. This is of course also preventing the docker run command from getting executed (which either way would not work if this fails
```
TASK [hammers : Copy kubeconfig.yml file] ******************************************************************************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: If you are using a module and expect the file to exist on the remote, see the remote_src option
fatal: [dev]: FAILED! => {"changed": false, "msg": "Could not find or access '/opt/site-config/node_custom_config/kubeconfig.yml' on the Ansible Controller.\nIf you are using a module and expect the file to exist on the remote, see the remote_src option"}
```

msherman64 (approved):
Looks good to me.

before merging, just make sure the commit message is formatted so that the subject line is separate from the body, see guides like https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

JOUNAIDSoufiane:
> Looks good to me.
> 
> before merging, just make sure the commit message is formatted so that the subject line is separate from the body, see guides like https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

Great, I will first verify that IP rotation works well before removing the dry run flag and merging

[EDIT] IP Rotation has been tested to work on CHI@Edge

PR #278, opened 2024-03-01
Added an if clause to zun.conf to disable RuntimeFilter if k8s backend is used

JOUNAIDSoufiane:


kolla/node_custom_config/zun.conf
```diff
@@ -25,7 +25,7 @@ heartbeat_in_pthread = false
 
 [scheduler]
 available_filters = zun.scheduler.filters.all_filters
-enabled_filters = {% if enable_blazar | bool %}BlazarFilter,{% endif %}ComputeFilter,RuntimeFilter
+enabled_filters = {% if enable_blazar | bool %}BlazarFilter,{% endif %}ComputeFilter,{% if not enable_zun_compute_k8s | bool %}RuntimeFilter{% endif %}
```
msherman64, line 28: making sure I remember the context, the runtimefilter is executed at the zun "host" level, and since we only have the one k8s "host" for zun, it breaks the nvidia runtime checks?

msherman64 (approved):
Just want to leave a commit message that details why this extra check is in there, so we're not confused coming back to it later.

kolla/node_custom_config/zun.conf
JOUNAIDSoufiane: well it breaks zun scheduling. zun just then says there are no more hosts

PR #280, opened 2024-03-05
Notify project manager about projects violating lease stacking policy

AnishReddyRavula:
Add lease stacking notifier

https://github.com/ChameleonCloud/hammers/pull/44

### DIFF
--- a/roles/hammers/templates/lease-stacking.json.j2
+++ b/roles/hammers/templates/lease-stacking.json.j2
@@ -10,6 +10,6 @@
     "min_nodes_for_coverage": 4,
     "exclude_node_types": [],
     "high_end_node_types": [],
-    "sender_email": "noreply@chameleoncloud.org",
-    "manager_email": "help@chameleoncloud.org"
+    "sender_email": "{{ hammer_lease_stacking_sender_email }}",
+    "manager_email": "{{ hammer_lease_stacking_manager_email }}"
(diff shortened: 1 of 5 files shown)


roles/hammers/templates/lease-stacking.json.j2
```diff
@@ -0,0 +1,15 @@
...
+        "openstack"
+    ],
+    "lease_coverage_threshold": 0.9,
+    "high_end_node_coverage_threshold": 0.5,
+    "minimum_lease_window_days": 21,
+    "min_nodes_for_coverage": 4,
+    "exclude_node_types": [],
+    "high_end_node_types": [],
+    "sender_email": "noreply@chameleoncloud.org",
+    "manager_email": "help@chameleoncloud.org"
```
msherman64: I would recommend making this settable via defaults.yml, like

```
"sender_email": "{{ hammer_lease_stacking_sender_email }}"
"manager_email": "{{ hammer_lease_stacking_manager_email }}"

I would also NOT default the manager_email to "help@chameleoncloud.org", due to the potential noise.
A manager can of course manually set it if necessary.

msherman64 (approved):
Mostly good, just a request to make a couple modifications to the config file to set the defaults.

codyhammock (commented):
LGTM

PR #281, opened 2024-04-01
dev-in-a-box with baremetal+tenks

msherman64:
This PR documents dev-in-a-box more fully, and provides a pre-configured defaults.yml file suitable to set up a development instance on top of KVM.

In particular, it uses 3 sets of veth pairs to create dummy interfaces for the external api, internal api, and neutron physnet, so that we don't depend on actual interface naming.


msherman64:
some assumptions for this to work:

a separate veth pair is created for each of:
* external_api
* internal_api
* physnet for public
* physnet for sharednet
* physnet for ironic-provisoning

the playbook attaches ironic-provisioning to the "first" tenks bridge/iface
and sharednet1 to the "second" bridge/iface.

Finally, you need to add the gateway IP for the ironic-provisioning network (whatever's configured in the subnet), to the brtenks0 bridge, so that nodes can route to the internal_api endpoint.

When you run the commands to "start up" tenks, it will create 3 libvirt domains, a virtualbmc listener for each, and a network interface for each VM attached to the provisioning and sharednet physnets.
Then, it will enroll them into ironic (but not blazar or Doni) for you.

to test things out, try commands like:
`openstack baremetal node inspect tk0` or `openstack server create --flavor my_rc --image CC-Ubuntu20.04 --network sharednet1 test01`

to watch the serial console, execute e.g. `virsh console tk0`, which you can exit by pressing `ctrl+a, ]`

note some remaining chi-in-a-box tech debt: it's still a bit clunky to import the cc-ubuntu images. Also, these fake nodes need something like 4GB of ram, mostly to unpack our centos8-based ironic initramfs. This could be minimized by using tinyipa instead.

codyhammock (approved):
👍🏻

msherman64:
closed in favor of https://github.com/ChameleonCloud/chi-in-a-box/pull/295
which has ovs+vlan support

PR #286, opened 2024-05-23
Feature/explicit updates

msherman64:
auto-updating chi-in-a-box has caused frequent usability issues with
long installs, broken upstream python dependencies, and uncertainty on
which kolla-ansible version is in use.

This commit refactors all dependency installs into an "update_deps"
command, which must be invoked before other cc-ansible commands.
update_deps can be run repeatedly to update an existing venv.

Finally, we no longer edit the virtualenv by adding a symlink from
$VIRTUALENV/share to the kolla-ansible base directoy. Instead, in
the for editable installs of kolla-ansible (used by default in ciab), we
look up the actual install path by referencing the python dist-info.

This approach matches the one used in kolla-ansible, and is a
prerequisite to support installing kolla-ansible from a local checkout.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "src/kolla-ansible"]
+	path = src/kolla-ansible
+	url = https://github.com/chameleoncloud/kolla-ansible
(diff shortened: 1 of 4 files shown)


codyhammock (approved):
Looks good to me. 👍🏻

PR #288, opened 2024-06-12
Soufianej k3s calicocrd fixx

JOUNAIDSoufiane:
Fixed calico global network policies so the command only executes if the policies are missing


msherman64 (approved):
LGTM, but please include a reference in the commit message to calico deprecating the use of kubectl apply for making these changes, as this is what motivated our change.

PR #289, opened 2024-06-12
Added conf option to scope wg hub to right netns

JOUNAIDSoufiane:
This change is to add the 'create_hub_in_root_netns' boolean option designed to scope the wireguard hub creation to the root network namespace if set to True. To set, add the option to the site's defaults.

### DIFF
--- a/kolla/defaults.yml
+++ b/kolla/defaults.yml
@@ -257,6 +257,7 @@ neutron_dev_plugins:
 
 # minimal config for neutron wireguard agent
 neutron_wireguard_hub_endpoint: "{{ kolla_external_vip_address }}"
+create_hub_in_root_netns: "{{ create_hub_in_root_netns }}"
 
 # Nova
 enable_nova: yes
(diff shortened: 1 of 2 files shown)


JOUNAIDSoufiane:
[Fix Chi@Edge Chi-in-a-box Wireguard hub creation in wrong namespace](https://trello.com/c/JDIXb2XU/3273-fix-chiedge-chi-in-a-box-wireguard-hub-creation-in-wrong-namespace)

msherman64:
Note! Requires code/container from this PR
https://github.com/ChameleonCloud/networking-wireguard/pull/5

kolla/defaults.yml
```diff
@@ -257,6 +257,7 @@ neutron_dev_plugins:
 
 # minimal config for neutron wireguard agent
 neutron_wireguard_hub_endpoint: "{{ kolla_external_vip_address }}"
+create_hub_in_root_netns: "{{ create_hub_in_root_netns }}"
```
msherman64: this needs a default:

normally I would say to default it to `false` so that just deploying the container won't change behavior.
however, as we're the only ones deploying this, lets set it to `true` because it will save some steps.

msherman64 (changes requested):
In the current version here, the ml2_conf.ini part is fine.

However, under `kolla/defaults.yml`, I suggest the following:

```
neutron_wireguard_create_hub_in_root_netns: True
```

JOUNAIDSoufiane:
Done

PR #290, opened 2024-06-12
New Task to create tunelo-calico network/subnet

JOUNAIDSoufiane:
Added a task in the k3s playbook to create the tunelo-calico network and the tunelo-calico subnet that are both necessary for the management of hub and spoke ports and the proper functioning of tunelo.

### DIFF
--- a/roles/k3s/tasks/config-neutron.yml
+++ b/roles/k3s/tasks/config-neutron.yml
@@ -52,6 +52,11 @@
 # FIXME(jason): this doesn't actually add the subnet to the router!
 # When we have updated to a later Ansible we can potentially fetch the router's
 # interfaces with routers_info and then merge this interface into the list?
+# Update (Soufiane June 25 2024) This still does not add the subnet to the router
+# two extra manual steps have to be performed after the succesful execution of this task
+# 1. Add the caliconet-subnet as an internal interface to the admin router
+# 2. Add an external gateway to the public network
+# Both of the above are pre-requisites to the proper functioning of floating ips to containers
 - name: Fetch existing NAT router
   kolla_toolbox:
     module_name: os_router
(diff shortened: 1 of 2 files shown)


JOUNAIDSoufiane:
[Fix CHI@Edge Chi-in-a-box not creating the tunelo-calico network which is required](https://trello.com/c/Fod5vbZs/3274-fix-chiedge-chi-in-a-box-not-creating-the-tunelo-calico-network-which-is-required)

roles/k3s/tasks/config-neutron.yml
```diff
@@ -23,6 +23,28 @@
       enable_dhcp: no
   become: yes
 
+- name: Create tunelo-calico network
+  kolla_toolbox:
+      module_name: os_network
+      module_args:
+        auth: "{{ openstack_auth }}"
```
msherman64, line 32: add `project: "{{ keystone_admin_project }}"`

roles/k3s/tasks/config-neutron.yml
```diff
@@ -23,6 +23,28 @@
...
+        name: tunelo-calico
+        provider_network_type: flat
+        provider_physical_network: tunelo
+        shared: false
+  become: yes
+
+- name: Create tunelo-calico subnet
+  kolla_toolbox:
+      module_name: os_subnet
```
msherman64, line 42: add `project: "{{ keystone_admin_project }}"`

msherman64 (changes requested):
I would leave the "{{ physical_network_calico }}" and "{{ physical_network_tunelo }}" hardcoded for now. 
I'll add templating for them when I refactor the post-deploy network creation, as it's not useful before then. (can't make two different flat networks on the same physnet, so need to template everything else)

Otherwise LGTM

PR #291, opened 2024-06-14
Soufianej taint k3s

JOUNAIDSoufiane:
## Added k8s worker taint configuration options
The options to add to site-config are:
- doni_enable_worker_taint: true/false, by default set to false (enables worker tainting through
  the k8s worker on doni devices)
- zun_tolerate_worker_taint: true/false, by default set to false (enables taint enforcement by
  adding tolerations to worker pods)
- k8s_worker_taint
  - key: taint key, by default set to "worker-node"
  - value: taint value, by default set to "true"
  - effect: taint effect, by default set to "NoSchedule"

## Added taint tolerations to core deployments in k3s
k3s defaults.yaml gets the value of k3s_worker_taint from worker_taint
which is defined under kolla/defaults.yml which subsequently defines
defaults for the option and takes in site values from the
k8s_worker_taint option that can be specified through the site config.

Added worker node taint toleration to smarter devices manager
daemonsets.

Furthermore, templated the nvidia device plugin daemonset and added the
toleration there as well.

Taint deployment strategy on a running testbed:

1. Redeploy k3s playbook to apply tolerations to and relaunch the core daemonsets running on the worker nodes
2. Set `zun_tolerate_worker_taint` to `True` and redeploy Zun
3. Finally, set `doni_enable_worker_taint` to `True` and redeploy 

The above sequence ensures that no existing or simultaneous user pods get evicted and inflicts minimal downtime to core daemonsets.

### DIFF
--- a/kolla/defaults.yml
+++ b/kolla/defaults.yml
@@ -345,6 +345,13 @@ enable_k3s: no
 enable_zun: no
 enable_zun_compute_k8s: "{{ enable_zun | bool and enable_k3s | bool }}"
 blazar_enable_device_plugin_k8s: "{{ enable_k3s | bool }}"
+# K8S worker tainting
+zun_tolerate_worker_taint: false
+doni_enable_worker_taint: false
+worker_taint:
+    key: "{{ k8s_worker_taint.key | default('worker-node') }}"
+    value: "{{ k8s_worker_taint.value | default('true') }}"
+    effect: "{{ k8s_worker_taint.effect | default('NoSchedule') }}"
 # If not running K3s, can run standard Zun deploy w/ etcd and kuryr
 enable_etcd: "{{ enable_zun | bool and not enable_k3s | bool }}"
 enable_kuryr: "{{ enable_zun | bool and not enable_k3s | bool }}"
--- a/roles/k3s/templates/nvidia-device-plugin.yaml.j2
+++ b/roles/k3s/templates/nvidia-device-plugin.yaml.j2
@@ -1,0 +1,74 @@
+# Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: nvidia-device-plugin-daemonset
+  namespace: kube-system
+spec:
+  selector:
+    matchLabels:
+      name: nvidia-device-plugin-ds
+  updateStrategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      # This annotation is deprecated. Kept here for backward compatibility
+      # See https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ""
+      labels:
+        name: nvidia-device-plugin-ds
+    spec:
+      nodeSelector:
+        "nvidia-device-plugin/enabled": "true"
+      tolerations:
+        # This toleration is deprecated. Kept here for backward compatibility
+        # See https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/
+        - key: CriticalAddonsOnly
+          operator: Exists
+        - key: nvidia.com/gpu
+          operator: Exists
+          effect: NoSchedule
+        - key: "{{ k3s_worker_taint.key }}"
+          operator: Equal
+          value: "{{ k3s_worker_taint.value }}"
+          effect: "{{ k3s_worker_taint.effect }}"
+      # Mark this pod as a critical add-on; when enabled, the critical add-on
+      # scheduler reserves resources for critical add-on pods so that they can
+      # be rescheduled after a failure.
+      # See https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/
+      priorityClassName: "system-node-critical"
+      containers:
+        - image: nvcr.io/nvidia/k8s-device-plugin:v0.14.1
+          name: nvidia-device-plugin-ctr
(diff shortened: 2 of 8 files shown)


JOUNAIDSoufiane:
[When deploying K3S. taint device nodes to only tolerate non-control plane services and add tolerations for device plugins and other device specifics.](https://trello.com/c/wtCFD1Yv/3276-when-deploying-k3s-taint-device-nodes-to-only-tolerate-non-control-plane-services-and-add-tolerations-for-device-plugins-and-oth)

msherman64:
Requires code from:

Zun: https://github.com/ChameleonCloud/zun/pull/20
Doni: https://github.com/ChameleonCloud/doni/pull/138


Order to apply is:
1. pull and deploy new zun
2. pull and deploy new doni
3. deploy this PR

msherman64:
note: add `worker_node_taint` with a `| default(something)` to kolla/defaults.yml

roles/k3s/templates/smarter-device-manager-ds-jetson.yaml.j2
```diff
@@ -69,6 +69,11 @@ spec:
       nodeSelector:
         smarter-device-manager: enabled
         smarter-device-manager/configmap: jetson
+      tolerations:
+        - key: "{{ k3s_worker_taint.split('=')[0] }}"
+        operator: "Equal"
+        value: "{{ k3s_worker_taint.split('=')[1].split(':')[0] }}"
+        effect: "{{ k3s_worker_taint.split(':')[1] }}"
```
msherman64: It is difficult to maintain jinja templates containing logic.

I suggest putting lines into kolla/defaults.yml instead:
```
k3s_worker_taint_key:
k3s_worker_taint_value:
k3s_worker_taint_effect:
```

If necessary, you can also use those to template the string format, also in defaults.yml
```
# format like 'worker-node=true:NoSchedule'
k3s_worker_taint: "{{ k3s_worker_taint_key }}={{ k3s_worker_taint_value }}:{{ k3s_worker_taint_effect}}"
```

roles/k3s/templates/smarter-device-manager-ds-rpi.yaml.j2
```diff
@@ -67,6 +67,11 @@ spec:
       nodeSelector:
         smarter-device-manager: enabled
         smarter-device-manager/configmap: rpi
+      tolerations:
+        - key: "{{ k3s_worker_taint.split('=')[0] }}"
+        operator: "Equal"
+        value: "{{ k3s_worker_taint.split('=')[1].split(':')[0] }}"
+        effect: "{{ k3s_worker_taint.split(':')[1] }}"
```
msherman64: suggest refactor as per above comment

roles/k3s/templates/nvidia-device-plugin.yaml.j2
```diff
@@ -0,0 +1,74 @@
...
+        # See https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/
+        - key: CriticalAddonsOnly
+          operator: Exists
+        - key: nvidia.com/gpu
+          operator: Exists
+          effect: NoSchedule
+        - key: "{{ k3s_worker_taint.split('=')[0] }}"
+          operator: "Equal"
+          value: "{{ k3s_worker_taint.split('=')[1].split(':')[0] }}"
+          effect: "{{ k3s_worker_taint.split(':')[1] }}"
```
msherman64: suggest refactor as per above comment

roles/k3s/templates/nvidia-device-plugin.yaml.j2
```diff
@@ -0,0 +1,74 @@
...
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop: ["ALL"]
+          volumeMounts:
+            - name: device-plugin
+              mountPath: /var/lib/kubelet/device-plugins
+      volumes:
+        - name: device-plugin
+          hostPath:
+            path: /var/lib/kubelet/device-plugins
```
msherman64, line 74: I would suggest breaking the "vendoring" of this template into a separate commit (same PR) as adding the taints/tolerations.

Otherwise this appears to be a very large change, and isn't immediately clear which lines are relevant to which intent

msherman64 (changes requested):
overall LGTM, but with the following caveats:

1. the noted refactoring of how the taint/toleration is specified, to avoid repeating logic throughout the jinja templates
2. related to the integration with doni and zun: I would suggest adding two flags:

`doni_enforce_worker_taints: false`
`zun_enforce_worker_taints: false`

Then, separate from the deployed container versions, the operator can first toggle on the zun support, and then toggle on the doni support, to avoid unexpected breakage.

Finally, is there a migration procedure to ensure currently running/deployed containers are not un-scheduled when this is deployed?
Do we have docs on how to trigger doni to deploy the new annotation? How can an operator check if some otherwise "good" nodes aren't available because they are/aren't tainted correctly?

kolla/defaults.yml
```diff
@@ -92,6 +92,7 @@ enable_cinder: no
 
 # Doni
 enable_doni: yes
+doni_enforce_worker_taints: false
```
msherman64: this seems unused?

kolla/defaults.yml
```diff
@@ -345,6 +346,10 @@ enable_k3s: no
 enable_zun: no
 enable_zun_compute_k8s: "{{ enable_zun | bool and enable_k3s | bool }}"
 blazar_enable_device_plugin_k8s: "{{ enable_k3s | bool }}"
+# K8S worker tainting
+zun_enforce_worker_taint: false
```
msherman64: should this be `zun_tolerate_worker_taint`? afaik it will not prevent scheduling even if no workers are tainted?

kolla/defaults.yml
```diff
@@ -345,6 +346,10 @@ enable_k3s: no
...
+doni_enable_worker_taint: false
+worker_taint: "{{ k8s_worker_taint.key | default('worker-node') }}={{ k8s_worker_taint.value | default('true') }}:{{ k8s_worker_taint.effect | default('NoSchedule') }}"
```
msherman64: Is this a string or a dict? you reference `worker_taint.key` in doni.conf?

msherman64 (commented):
Added some comments, mostly just naming/nits for consistency and to avoid confusing future me?

kolla/defaults.yml
JOUNAIDSoufiane: remnant from a merge :(, I'll change it

kolla/defaults.yml
JOUNAIDSoufiane: agreed

kolla/defaults.yml
JOUNAIDSoufiane: this is a dict, looks like my merge was a little destructive

PR #292, opened 2024-06-19
Soufianej k3s playbook

JOUNAIDSoufiane:
- Removed the ansible_managed flag from k3s template files
- Added a pip task to install the kubernetes python library

### DIFF
--- a/roles/k3s/tasks/config-k3s-service.yml
+++ b/roles/k3s/tasks/config-k3s-service.yml
@@ -65,4 +65,5 @@
   become: true
   pip:
     name: kubernetes
+    virtualenv: "{{ virtualenv }}"
     state: present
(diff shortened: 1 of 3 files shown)


roles/k3s/tasks/config-k3s-service.yml
```diff
@@ -60,3 +60,9 @@
     daemon_reload: yes
     state: started
     enabled: yes
+
+- name: Install Kubernetes Python client
+  become: true
+  pip:
+    name: kubernetes
+    state: present
```
msherman64: Did you verify that this installed client into `/etc/ansible/venv`, such as by deleting the venv, then re-creating with bootstrap-servers prior to running this playbook?

Also, we should pin the version of python-kubernetes installed, or we might get unexpected breakage later

msherman64 (approved):
LGTM, assuming nits are double-checked

roles/k3s/tasks/config-k3s-service.yml
JOUNAIDSoufiane: sure looking into

codyhammock (approved):
LGTM

PR #297, opened 2024-08-05
refactor mysql s3 backup task

msherman64:
The primary change in this commit is to replace use of the "aws-cli"
tool with "rclone", as the AWS tools are not compatible with ceph rgw
based s3 implementations.

Additionally, this removes support for the mysqldump based legacy
backups, and refactors the backup script for readability and improved
logging.

Behavior changes:
Rather than only the latest backup file, rclone will copy any new files
that are not present in the object store, but will NOT overwrite
existing ones.

### DIFF
--- a/roles/chameleon_mariadb/defaults/main.yml
+++ b/roles/chameleon_mariadb/defaults/main.yml
@@ -5,11 +5,17 @@ deployment_dir: "{{ deployment_dir }}"
 deployment_site_config_dir: "{{ site_config_dir }}"
 mariadb_backup_timer_suffix: "{{ site_config_dir | basename }}"
 mariadb_backup_cron_script: /usr/local/sbin/mariadb-backup
-mariadb_backup_max_packet: 1M
 mariadb_backup_file_age: 29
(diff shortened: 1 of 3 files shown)


msherman64:
One concern (not sure best way to fix) is how to disable this.

We'd basically need a "enable_mariadb_backup_auto" flag, and have ansible ensure the relevant scripts and timers are removed if not set to true.

Mark-Powers:
What is actually creating the backups now? Is this syncing the volume data directly?

msherman64:
> What is actually creating the backups now? Is this syncing the volume data directly?

The backups are created by a cronjob on the deploy host which executes `cc-ansible mariadb_backup`, which uses mariabackup to create backups from the running mariadb server/cluster, and stores those backups into the `mariadb_backup` docker volume.

the cron job here just handles rotating those backup files, and copying them from this docker volume to an object store bucket.

codyhammock (approved):
Looks good! 👍🏻

Mark-Powers (approved):

PR #298, opened 2024-08-05
xena: [db backups] remove mysqldump, use rclone for s3

msherman64:
Backport of https://github.com/ChameleonCloud/chi-in-a-box/pull/297

The primary change in this commit is to replace use of the "aws-cli" tool with "rclone", as the AWS tools are not compatible with ceph rgw based s3 implementations.

Additionally, this removes support for the mysqldump based legacy backups, and refactors the backup script for readability and improved logging.

Behavior changes:
Rather than only the latest backup file, rclone will copy any new files that are not present in the object store, but will NOT overwrite existing ones.


codyhammock (approved):
Looks good! 👍🏻

PR #301, opened 2024-08-22
add encrypt_file/decrypt_file methods

msherman64:
Kolla ansible supports automatic decryption of vault-encrypted files in site-config.
This is very useful for things like ceph authentication tokens, certificates, and so on.

Note: This CANNOT be used for the SSH key for ansible to connect to a remote host.
          See https://github.com/ansible/ansible/issues/22382

Add cc-ansible wrappers for `vault encrypt` and `vault decrypt` to make this easy.


codyhammock (approved):
LGTM. Do the github-actions flags need to be addressed?

PR #302, opened 2024-08-22
move ansible_python_interpereter to all.yaml so it can be overridden

msherman64:
This is going to need an upgrade note, since it will require modification of the site-config or things will break.


msherman64:
This is going to need an upgrade node, since it will require modification of the site-config or things will break.

codyhammock (approved):
LGTM. I'd also be good with requiring /etc/ansible/venv be created on the deploy host, as it may assist with compartmentalization

PR #304, opened 2024-09-04
update ironic kernel append params

msherman64:
this PR has a couple of changes:

1. in ironic, `pxe_append_params` is deprecated, so we need to replace it with `kernel_append_params`
2. the `nodemodeset` and `systemd.journald.forward_to_console` options can both cause blocking and performance issues, so they have been removed.

Links to issues:
https://github.com/openstack/ironic/blob/e01522cd4ca02897bbecc4190ce128c5a344fd9f/releasenotes/notes/remove-nomodset-7a352a9519c1045b.yaml#L5


codyhammock (approved):
It's a straightforward update 👍🏻

PR #314, opened 2025-01-02
ensure ceph rgw port defaults to 7480

### LINKED ISSUE #313: ceph rgw port is not applied
During the antelope upgrade, the config parameter to set the port for ceph rgw was changed.
we need to update `rgw_port` to `ceph_rgw_port`.

msherman64:
our ceph rgws listen on port 7480, and we need to configure the keystone endpoint to match. We had been doing this in the chi-in-a-box defaults via `rgw_port`, but the kolla syntax has changed to `ceph_rgw_port` instead.

A separate question is should we be overriding this at all anymore, but that's out of scope.


codyhammock (approved):
simple change; LGTM

PR #317, opened 2025-01-24
reduce memory usage

msherman64:
This PR reduces memory usage by memcached and mariadb.

However, we may want to put more effort into the new default values used, and/or make it more easily configurable by the site operator.


codyhammock (approved):
LGTM

Mark-Powers (approved):
Looks good, I agree regarding easier configuration of these values, but this seems fine as default.

PR #318, opened 2025-01-27
Add flavor reservation configuration

Mark-Powers:
This adds configuration needed for the blazar flavor plugin.

### DIFF
--- a/kolla/defaults.yml
+++ b/kolla/defaults.yml
@@ -372,6 +372,8 @@ enable_image_cleaner: no
 default_share_type: cephfsnfstype
 
 # Flavor reservation
-enable_host_reservation: "{{ enable_nova | bool }}"
-enable_flavor_reservation: no
+blazar_enable_host_reservation: "{{ enable_nova | bool }}"
+blazar_enable_flavor_reservation: no
 blazar_randomize_hosts: no
+# The trait which permits flavor reservation
+blazar_flavor_reservation_trait: blazar_flavor_reservation
--- a/kolla/defaults.yml
+++ b/kolla/defaults.yml
@@ -370,3 +370,8 @@ enable_image_cleaner: no
 
 # Manila
 default_share_type: cephfsnfstype
+
+# Flavor reservation
+# introducing new value here to enable host plugin. Before we were default to `enable_nova`
+enable_host_reservation: "{{ enable_nova | bool }}"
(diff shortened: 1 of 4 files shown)


Mark-Powers:
@msherman64 Is there a way I can assert a condition on the defaults? I want to ensure
`enable_host_reservation nand enable_flavor_reservation`

msherman64:
> @msherman64 Is there a way I can assert a condition on the defaults? I want to ensure `enable_host_reservation nand enable_flavor_reservation`

there's not really a way to do this within the defaults file / config file. You either need to do this in an ansible task (like in prechecks or config steps, in kolla-ansible), or via an external tool like ansible-lint. In chi-in-a-box itself, you're almost limited to the logic that jinja can do.

will blazar fail if both host_reservation and flavor_reservation are enabled?

Mark-Powers:
Main issue is that our host_reservation code assumes a lot about hosts, and the monitoring fails for some reason. I'm still digging into this.

kolla/defaults.yml
```diff
@@ -370,3 +370,8 @@ enable_image_cleaner: no
 
 # Manila
 default_share_type: cephfsnfstype
+
+# Flavor reservation
+enable_host_reservation: "{{ enable_nova | bool }}"
```
codyhammock: I think that it be more clear and consistent to name these labels with "blazar_", e.g. "blazar_enable_host_reservation".

msherman64 (changes requested):
Seconding the request to prefix the new variables with `blazar_`, due to the global namespace.

Additionally, we seem to maintain a list of plugins and their state in kolla-ansible's blazar role: 
https://github.com/ChameleonCloud/kolla-ansible/blob/568df787eeba52116c081ba575c60152242d7135/ansible/roles/blazar/defaults/main.yml#L42-L57

And that lets us template the list of plugins in the config file more cleanly:
https://github.com/ChameleonCloud/kolla-ansible/blob/568df787eeba52116c081ba575c60152242d7135/ansible/roles/blazar/defaults/main.yml#L62
https://github.com/ChameleonCloud/kolla-ansible/blob/568df787eeba52116c081ba575c60152242d7135/ansible/roles/blazar/templates/blazar.conf.j2#L20

msherman64:
I believe this can now be closed.

PR #326, opened 2025-03-10
Add blazar cleaning time and host randomization config

Mark-Powers:


msherman64 (approved):
this looks good to me, but is there somewhere we document the new config settings for blazar, so that it's possible to go back and look up what these settings mean?

Mark-Powers:
These settings are from upstream blazar, and so we can refer to openstack's documentation https://docs.openstack.org/blazar/latest/configuration/blazar-conf.html#default

msherman64:
> These settings are from upstream blazar, and so we can refer to openstack's documentation https://docs.openstack.org/blazar/latest/configuration/blazar-conf.html#default

Ah, sounds good, makes it clear it's not just an "us" thing.

PR #335, opened 2025-03-13
Minimal/2023.1 upstream

msherman64:
this is just a placeholder PR, this holds the delta between our current stable branch, and the minimum necessary to deploy a kvm site via kolla-ansible (with no CHI features)


msherman64:
Closing, we finished the upgrade.

PR #341, opened 2025-04-18
Add glance property protections for chameleon-supported metadata

pdmars:
I'm not sure on some of my selections here, feedback welcome and needed


kolla/defaults.yml
```diff
@@ -108,6 +108,8 @@ glance_backends:
     enabled: "{{ glance_backend_swift | bool }}"
 
 glance_allow_anonymous_access: no
+glance_property_protections_file: "{{ node_custom_config }}/glance/property_protections.conf"
```
pdmars: is this the correct way to make sure we get this property_protections.conf file propagated where we need it?

kolla/defaults.yml
```diff
@@ -108,6 +108,8 @@ glance_backends:
...
+glance_openstack_user: "{{ keystone_admin_user }}"
```
pdmars: is this the way to reference the admin user we want to restrict this metadata field to?

kolla/node_custom_config/glance/property_protections.conf
```diff
@@ -0,0 +1,12 @@
...
+read   = @ # anyone can read
+update = {{ glance_openstack_user }}
+delete = {{ glance_openstack_user }}
+
+# catch‑all for any other properties – allow
+[.*]
+create = @
+read   = @
+update = @
+delete = @
```
msherman64: this file should get put in the path `kolla/node_custom_config/glance/property-protections-rules.conf` so that the corresponding kolla-ansible task picks it up:

https://github.com/ChameleonCloud/kolla-ansible/blob/chameleoncloud/2023.1/ansible/roles/glance/tasks/config.yml#L141-L154

kolla/node_custom_config/glance.conf
```diff
@@ -1,5 +1,7 @@
 [DEFAULT]
 allow_anonymous_access = {{ glance_allow_anonymous_access }}
+property_protection_file = {{ glance_property_protections_file }}
```
msherman64: this can be left out, kolla-ansible already defines the following in the glance-api template:
https://github.com/ChameleonCloud/kolla-ansible/blob/568df787eeba52116c081ba575c60152242d7135/ansible/roles/glance/templates/glance-api.conf.j2#L21-L23

this path refers to the config file path where the glance-api service is running, *inside* the container.

kolla/node_custom_config/glance/property_protections.conf
```diff
@@ -0,0 +1,12 @@
+[chameleon-supported]
```
msherman64: since the section header can be a regex, maybe we want to define a namespace of sorts?
`chameleon.supported` or whatever?, and later have chameleon.deprecated, chameleon.fancy,....

kolla/node_custom_config/glance/property_protections.conf
msherman64: good!

kolla/node_custom_config/glance/property_protections.conf
```diff
@@ -0,0 +1,12 @@
...
+create = {{ glance_openstack_user }}
+read   = @ # anyone can read
+update = {{ glance_openstack_user }}
+delete = {{ glance_openstack_user }}
```
msherman64: with the format==roles, this must correspond to an existing keystone user role. To restrict it to admins, we'd set this to `admin`.
however, I'd prefer to use policies, since we already configure policy rules, and the syntax is more useful than for 
roles. 

In order of specifcity, see:
- https://opendev.org/openstack/glance/src/branch/unmaintained/2023.1/etc/property-protections-policies.conf.sample
- https://docs.openstack.org/glance/2023.1/configuration/glance_policy.html
- https://docs.openstack.org/kolla-ansible/latest/admin/advanced-configuration.html#openstack-policy-customisation
- https://docs.openstack.org/security-guide/identity/policies.html

for now, I'd probably use policy `context_is_admin` for the supported tags, but we leave the door open to later define "project X in namespace Y owns supported images, so admins and/or users with member role on project X can modify these tags"

for the default case, I *think* we can use the policy named `default`, but may need to define a quick `owner_or_admin` policy, which we've done elsewhere

kolla/defaults.yml
msherman64: if we specify a policy, this can be left out.

worth revisiting later if we want more customization. For now using the "default admin user" is what we do implicitly

msherman64 (changes requested):
so, while I left a flood of comments, this is pretty close

I'd suggest the following:
1. rename the policy file so KA will pick it up
2. use policy, rather than "roles" format in glance-api.conf
3. in kolla/defaults.yml, set `glance_enable_property_protection: true`, so that the kolla-ansible jinja template adds the `property_protection_file=...` line

To test this, I'd use the PR branch of chi-in-a-box to deploy dev-in-a-box, since that will allow using the customized horizon + customized glance.

the isolated glance config file behavior could be tested via devstack, but you'd need to redo the setup with dev-in-a-box for horizon anyway.

kolla/node_custom_config/glance/property_protections.conf
pdmars: we could definitely make it so this rule applies to any `[chameleon-.*]` metadata. I'd worry slightly that that may break anyone who could be using chameleon- but the risk is probably pretty low. otherwise we can add `chameleon-deprecated` or whatever else we want as needed. I'm fine with either approach.

pdmars:
@msherman64 ok I think I addressed all of your comments. I was also able to test it and it appears to be working as expected.

msherman64 (approved):
looks good to me then!

were you able to view (and confirm not editing) the properties as a non-admin user?
that's all I can think of before merging.

pdmars:
> looks good to me then!
> 
> were you able to view (and confirm not editing) the properties as a non-admin user? that's all I can think of before merging.

Yep! I inspected the glance setup and all looked good. Also created a non-admin user and was able to verify that I could modify other metadata, but not the protected one:

```
$ openstack image set --property new_field=false 516640b3-204b-4e5f-bd13-1f4aa6b1a968
$ openstack image show 516640b3-204b-4e5f-bd13-1f4aa6b1a968
+------------------+------------------------------------------------------------------+
| Field            | Value                                                            |
+------------------+------------------------------------------------------------------+
| container_format | bare                                                             |
| created_at       | 2025-04-08T16:40:04Z                                             |
| disk_format      | raw                                                              |
| file             | /v2/images/516640b3-204b-4e5f-bd13-1f4aa6b1a968/file             |
| id               | 516640b3-204b-4e5f-bd13-1f4aa6b1a968                             |
| min_disk         | 0                                                                |
| min_ram          | 0                                                                |
| name             | CC-Ubuntu22.04                                                   |
| owner            | 7e44d885d37647f5ab152f1ff348bf21                                 |
| properties       | chameleon-supported='true', os_hidden='False', new_field='false' |
| protected        | False                                                            |
| schema           | /v2/schemas/image                                                |
| status           | queued                                                           |
| tags             |                                                                  |
| updated_at       | 2025-04-21T21:29:39Z                                             |
| visibility       | public                                                           |
+------------------+------------------------------------------------------------------+
```

And the protected one:
```
$ openstack image set --property chameleon-supported=false 516640b3-204b-4e5f-bd13-1f4aa6b1a968
ForbiddenException: 403: Client Error for url: http://172.16.200.254:9292/v2/images/516640b3-204b-4e5f-bd13-1f4aa6b1a968, Forbidden
```

codyhammock (approved):
LGTM. Good test 👍🏻

msherman64:
This looks great. I've just added a test for this as well: https://github.com/ChameleonCloud/chi-blazar-tempest-plugin/pull/4
and that will let us verify it's enabled, and at which sites.

PR #342, opened 2025-04-29
[ironic] disable node history due to bug 2054594

msherman64:
Workaround for nodes getting stuck in "bad" states until restart of ironic conductor, as well as recovery from network-related errors.

LP: https://bugs.launchpad.net/ironic/+bug/2054594 ironic uses a 32 character DB column for user_id, which is too short for the 64 character ids returned by keystone with federated login

the real fix is a db migration, but that is hugely painful to try and backport from 2024.1. Instead, just diable the node_history feature, which seems to be the only thing in ironic writing 64 character IDs.

Revert after upgrading to 2024.1.


codyhammock (approved):
No notes

PR #344, opened 2025-05-22
Replace old image deployer with new one

pdmars:


Mark-Powers (approved):
Looks good!

kolla/defaults.yml
```diff
@@ -369,6 +369,11 @@ enable_usage_reporting: no
 # image tools
 enable_image_deployer: no
 enable_image_cleaner: no
+image_container: chameleon-supported-images
+image_type: qcow2
+image_prefix: testing_
+image_scope: prod
+image_object_store_url: https://chi.tacc.chameleoncloud.org:7480/swift/v1/AUTH_key
```
msherman64: shouldn't this be:
```
image_object_store_url: https://chi.tacc.chameleoncloud.org:7480/swift/v1/AUTH_f6c7696906c04b3c89fc3bda9a1b8be0
```

roles/chameleon_image_tools/tasks/main.yml
```diff
@@ -6,6 +6,58 @@
...
+  ignore_errors: yes
+  when: enable_image_deployer | bool
+
+- name: Remove old wrapper script
+  become: yes
+  file:
+    path: /usr/local/sbin/chameleon-image-deploy
+    state: absent
+  ignore_errors: yes
+  when: enable_image_deployer | bool
```
msherman64: these should be moved to a block or separate task file

and, instead of ignoring errors, which still print in red, we should have a step which checks for the presence of the old systemd units, and IFF that check finds them, runs the removal tasks

msherman64 (changes requested):
with this playbook as-is, authentication and image download don't seem to be working.
I've added a few commits to the end to fix this.

kolla/defaults.yml
pdmars: Yeah, I missed updating it later after testing that locally.

msherman64:
Does this work now if you start with an empty `/etc/hammers` directory? I wasn't sure how authentication had been working previously.

docs/operations/chameleon-tools/image-tools.md
```diff
@@ -26,46 +26,14 @@ By default, the automatic deployment task is _disabled_. To enable the task, set
...
-* Delete images that are 18 months or older. You can also tune this number by overwriting `delete_image_age_in_month` in the site `defaults.yml`.
-
-By default, the automatic cleaning task is _disabled_. To enable the task, set `enable_image_cleaner: yes` in the site `defaults.yml`. Then, run the `post-deploy` or run the `chameleon_image_tools` playbook. The automatic cleaning task runs daily to hide or delete the older version and deprecated Chameleon-supported images. You can check the installed automatic cleaning task via the systemd [timers](https://wiki.archlinux.org/index.php/Systemd/Timers).
-
-```shell
-systemctl list-timers 'image_clean'
+docker run --rm --net=host -v "/etc/hammers:/etc/hammers:ro" \
+ghcr.io/chameleoncloud/chameleon_site_tools:sha-583d81d \
+image_deployer --site /etc/hammers/site.yaml --dry-run
 ```
```
msherman64: I think the `docker run` invocation can be replaced with `systemctl start chameleon-image-deploy.service`

pdmars:
> Does this work now if you start with an empty `/etc/hammers` directory? I wasn't sure how authentication had been working previously.

Yeah it did work for me, at least on my dev-in-a-box setup.

docs/operations/chameleon-tools/image-tools.md
pdmars: done

msherman64 (approved):
Thanks for the changes, looks good!

PR #349, opened 2025-05-30
fix glance +nova policy to workaround app credential bug

msherman64:
see https://bugs.launchpad.net/keystone/+bug/2030061

in keystone, the member role implies the reader role
and glance's default policy looks for `role:admin or (role:reader and project_id:%(project_id)s)`

at some point between train and antelope
keystone's handling of implicit role assignments broke for app credentials
and glance's default policy changed to use the role reader instead of member
and we don't see it on admin accounts because the admin role is not properly scoped, see https://bugs.launchpad.net/keystone/+bug/968696

and therefore glance uses our admin role even though it's set on the openstack project, and not the one in question

as a workaround, override all places where glance policy specifies `role:reader`, and replace with a rule that can be `role:member` or `role:reader`


codyhammock (approved):
LGTM 👍🏻

Mark-Powers (approved):

msherman64:
note: if this works, we'll probably also need to cherry-pick it and apply to baremetal sites

msherman64:
Note: changes can be tested by creating an application credential with _only_ the member role (via openstack application credential create --role member) and seeing if that credential permits `openstack image list`

changes were tested via dev-in-a-box, but we don't yet have smoke tests for this issue.

PR #352, opened 2025-08-27
Add systemd timer for periodic inspector

pdmars:


roles/chameleon_periodic_inspector/templates/periodic-inspector.service.j2
```diff
@@ -0,0 +1,20 @@
...
+ExecStart=/usr/bin/docker run --rm --net=host \
+    --env-file {{ chameleon_resource_discovery_config_dir }}/envvars \
+    {{ chameleon_resource_discovery_docker_image }} \
+    periodic_inspector \
+        --parallel {{ chameleon_resource_discovery_parallel }} \
+        --limit {{ chameleon_resource_discovery_limit }} \
+        {% if chameleon_resource_discovery_provide_manageable %}--provide-manageable {% endif %}\
+        {% if chameleon_resource_discovery_reinspect_failed %}--reinspect-failed {% endif %}\
+        --expire-days {{ chameleon_resource_discovery_expire_days }} \
+        --cloud {{ cloud }}
```
msherman64: if we're passing auth_url and so on  via env vars, I believe this should be the string `envvars`

roles/chameleon_periodic_inspector/templates/periodic-inspector.timer.j2
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Run Chameleon Periodic Inspector daily
+
+[Timer]
+OnCalendar=daily
```
msherman64, line 5: we'll probably want to play with this timing in practice
running more frequently allows the script to attempt to "fix" nodes that failed to inspect a previous time.

site-config.example/inventory/hosts
```diff
@@ -842,6 +842,10 @@ control
 [chameleon-image-tools:children]
 control
 
+# Periodic Inspector
+[chameleon-periodic-inspector:children]
+control
```
msherman64, line 847: all this script needs is access to the ironic API.

in practice "control" is fine, longer term we may want explicitly target the "deployment" group, and/or something like "control[0]" to avoid running it multiple times in the multinode case.

tldr leaving it as "control" is ok because we'll have to fix all sorts of other things before that assumption becomes invalid.

msherman64 (approved):
Looks great to me, thanks!

FYI, this can be tested without the timer by executing `systemctl start <unitname>` instead of having the timer do it.

Once deployed, we'll also want to remove my scripts from the "rally-executor" vm so that we're not doubling up on it

msherman64 (approved):

PR #354, opened 2025-09-04
add templates for central monitoring

msherman64:
This takes our current, ad-hoc configuration for sending metrics+logs to our central infra, and finally merges it into chi-in-a-box.

Two new toggles are added, defaulting to false, `upload_logs_chameleon` and `upload_metrics_chameleon`.
The fluentd and prometheus template files are only populated if these are true, preventing the current behavior from changing.

This also assumes that `keystone_idp_client_id` is used as the username for each site for pushing to both loki and mimir, and that the `prometheus_remote_write_password` is used as the password for both.

### DIFF
--- a/kolla/defaults.yml
+++ b/kolla/defaults.yml
@@ -331,13 +331,18 @@ enable_swift: no
 enable_swift_rgw: no
 
 # Monitoring
-# TODO: install ES curator
-enable_central_logging: yes
+enable_central_logging: no
 es_heap_size: 8G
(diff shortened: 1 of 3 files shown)


pdmars (approved):

msherman64:
Of note, any sites that already have these files in their site config will:
- (if the names match) override this config
- (if the names don't match) probably duplicate the config, but not unless the "upload_" toggles are turned on

the migration path is to just remove the overrides from site config once these is merged and enabled

PR #364, opened 2026-02-28
Feature/merge kvm branch 2023.1

msherman64:
This is an attempt to cleanly merge all of the kvm-specific changes into stable/2023.1 so we can go back to having one release branch.

it needs testing before merge.

### DIFF
--- a/kolla/defaults.yml
+++ b/kolla/defaults.yml
@@ -90,6 +90,8 @@ blazar_network_polling_monitor: true
 blazar_network_polling_monitor_dry_run: true
 blazar_randomize_hosts: true
 blazar_cleaning_time_minutes: 5
+blazar_filter_vm_hosts: true
+blazar_filter_ironic_hosts: true
 
 # enable to use https://github.com/ChameleonCloud/nova/pull/9
 # workaround for https://bugs.launchpad.net/nova/+bug/1542491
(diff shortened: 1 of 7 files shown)


msherman64:
So while this needs review, it was my attempt at capturing all the changes we make to make kvm work with chi-in-a-box.

My proposal is that we can pull single, or small sets of commits off of here via cherry-picks into their own PRs until we end up with one branch for both baremetal and kvm.

kolla/defaults.yml
```diff
@@ -47,7 +47,34 @@ backup_location: "{{ mariabackup_location if enable_mariabackup | bool else mysq
...
 blazar_project_enforcement_id: charge_code
+
+# override value in kolla-ansible/ansible/roles/blazar/defaults/main.yml
+blazar_enable_plugin_network: "{{ blazar_enable_network_reservation | bool }}"
+
+# overrides dict in kolla-ansible/ansible/roles/blazar/defaults/main.yml
+blazar_plugins:
+  - name: "physical.host"
+    enabled: "{{ enable_nova | bool }}"
+  - name: "virtual.instance"
```
Mark-Powers, line 67: `virtual.instance` should never be enabled for us

Mark-Powers (commented):
Looks good for the most part, but we probably want to always disable the virtual.instance plugin.

pdmars (approved):
Looks good to me pending testing. Are there any docs we should make sure to update that describe how/where/what branches all this stuff lives on and is deployed from?

PR #368, opened 2026-03-18
add more audit tables for cloud traces

msherman64:
Starting with blazar extra capabilitites and resource_properties
this should ensure we can track how the properties change over time for each host.

### DIFF
--- a/roles/audit_tables/defaults/main.yml
+++ b/roles/audit_tables/defaults/main.yml
@@ -4,6 +4,9 @@ audit_blazar_computehosts: false
 audit_blazar_computehost_extra_capabilities: "{{ audit_blazar_computehosts }}"
 audit_blazar_resource_properties: "{{ audit_blazar_computehosts }}"
 
+audit_nova_compute_nodes: false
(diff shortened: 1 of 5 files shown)


Mark-Powers (approved):
Looks good, but will double check the trace requirements later

roles/audit_tables/defaults/main.yml
```diff
@@ -1,6 +1,13 @@
 ---
 
 audit_blazar_computehosts: false
+audit_blazar_computehost_extra_capabilities: "{{ audit_blazar_computehosts }}"
+audit_blazar_resource_properties: "{{ audit_blazar_computehosts }}"
+
+audit_nova_compute_nodes: false
+audit_nova_services: "{{ audit_nova_compute_nodes }}"
+
+audit_ironic_nodes: false
```
msherman64, line 10: @codyhammock FYI, all of these default to disabled, so there should be no risk in deployments
you'll end up needing to opt in to each one, and e.g. don't enable ironic on kvm

PR #369, opened 2026-03-23
add nova instances table to audit

msherman64:


roles/audit_tables/defaults/main.yml
```diff
@@ -133,6 +134,79 @@ audit_tables:
...
+      - node
+      - deleted
+      - locked_by
+      - cleaned
+      - ephemeral_key_uuid
+      - hidden
+    watch_columns:
+      - power_state
+      - vm_state
+      - task_state
```
msherman64, line 201: this will generate a bunch of rows as an instance is created and provisioned, and also when instance actions happen

msherman64:
we could also add the instance_actions and instance_actions_events tables here, but the relevant info may also be captured by just the audit timestamps for the instances table.

roles/audit_tables/defaults/main.yml
```diff
@@ -207,6 +209,51 @@ audit_tables:
...
+      - finish_time
+      - result
+      # - traceback
+      - deleted
+      - host
+      # - details
+    watch_columns:
+      - finish_time
+      - result
+      - host
```
msherman64, line 256: for the most part, we'll end up with two audit rows per source row here, once when it's created, and one when the event finishes.

none of these show as "deleted", but we'd get a row for that too.

Mark-Powers (approved):

PR #372, opened 2026-03-26
Feature/audit leases and allocations

msherman64:
@Mark-Powers do we need any other columns/tables to capture things for flavor reservations?


Mark-Powers (approved):
This should be fine. A flavor reservation translates into an instance reservation, which caches the original info in resource_properties, and a host allocation.

PR #374, opened 2026-03-26
Add site name to site specific services/timers

Mark-Powers:
This adds the site name to the systemd services/timers rather than using names that may collide if a deploy host is used for multiple sites. I'm not 100% sure if this covers everything, but the other systemd services seemed like central services, rather than site specific ones.

### DIFF
--- a/roles/chameleon_image_tools/tasks/main.yml
+++ b/roles/chameleon_image_tools/tasks/main.yml
@@ -39,7 +39,7 @@
   become: yes
   template:
     src: chameleon-image-deploy.service.j2
-    dest: /etc/systemd/system/chameleon-image-deploy.service
+    dest: /etc/systemd/system/{{ chameleon_site_name }}-chameleon-image-deploy.service
     mode: '0644'
   when: enable_image_deployer | bool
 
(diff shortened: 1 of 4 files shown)


pdmars (approved):

msherman64:
@codyhammock can you check that chameleon_site_name is actually unique between chi@tacc and kvm?

I know there are a bunch of overlapping "name of site" variables, wasn't sure about this one in particular.

msherman64 (approved):

Mark-Powers:
@msherman64 Will this require manual cleanup before we can deploy? I expect it will create a duplicate timer.

msherman64:
> @msherman64 Will this require manual cleanup before we can deploy? I expect it will create a duplicate timer.

it will, this has a "template files" task, but no "delete files".

If any of these tasks create config directories/files, we'll need to check that they also don't conflict (like /etc/chameleon_usage)

codyhammock (approved):

msherman64:
as a sanity check @codyhammock 

on staff.chameleon.tacc.utexas.edu, the only hammer/periodic task I see a timer for is `mariadb_backup_site-config.timer`, which should be configurable via `mariadb_backup_timer_suffix` (defaults to name of site_config dir)

all the others deploy to a controller node, so they should not overlap per site?

codyhammock:
@msherman64 Yes, that appears to be correct.

PR #379, opened 2026-03-27
Feature/unit tests

msherman64:
This mostly adds a framework for running and extending tests, using pytest for fixtures and then having it run e.g. yq, or kolla-ansible genconfig.

For the fast "unit" tests, this is just exercising the config merge that cc-ansible does with yq + rsync, and then checking some basic invariants about chi-in-a-box, such as "all groups in kolla-ansible/inventory/multinode should show up in site-config.example" and "all entries in kolla-ansible/etc/passwords.yml should show up in site-config.example"


msherman64:
This already turned up a couple of easy fixes, such as tunelo missing from the site-config.example groups, and finding that the pinned kolla-ansible submodule was out of date.

PR #380, opened 2026-03-27
Feature/integration tests

msherman64:
Add tests to exercise "install_deps" and "genconfig"

Pytest fixtures allow passing composable sets of "defaults.yml" entries, then parsing the templated ini output files and asserting on the contents.

The goal is that we'll be able to say "chi-in-a-box supports specifying pci passthrough with syntax X"
And if we break a template, or kolla-ansible drops a default var, or something else goes wrong, the test breaks.

Hopefully this will help prevent surprises when doing things like merging KVM and Baremetal, without incurring too much maintenance overhead.


msherman64:
Even just running "install_deps" was a win, identified a couple of breakages.

PR #399, opened 2026-05-23
baremetal: skip disk size check for images

msherman64:
If users attempted to use an image larger than 20GB, they would receive a 400 error from nova, as the image size exceeded the flavor's disk. This also caused the default partion size to limit to 20gb.

This caused issues most frequently when users want to install large libraries (see CUDA, or FPGA libs), or when using RAW vs QCOW images, as nova's check looks at the file size too.

Sets the flavor's disk size to 0 gb, skipping the nova check, and set os_compute_api:servers:create:zero_disk_flavor so non-admins can use it. Policy prevented scheduler issues for KVM, but not relevant for baremetal. See https://docs.openstack.org/nova/2023.1/user/flavors.html#overview

Note: If an image is too large to fit on a baremetal node, users will see a failure when ironic tries to write the disk, instead of when submitting the POST to nova, which may delay feedback.

### DIFF
--- a/docs/operations/troubleshooting/known-issues/ironic-bare-metal.md
+++ b/docs/operations/troubleshooting/known-issues/ironic-bare-metal.md
@@ -16,4 +16,3 @@ This can happen for _many_ reasons. Check a few things:
 
 * Are there errors in the `/var/log/ironic/ironic-conductor.log`?
 * Is the node in maintenance mode? (It must be set to the "available" status for Nova to consider it.)
-* Does the node have enough space on the file system? It must have more than the space defined in the `baremetal` Nova flavor (we set this to a low value on purpose; 20Gb).
--- a/kolla/node_custom_config/nova/policy.yaml
+++ b/kolla/node_custom_config/nova/policy.yaml
@@ -60,3 +60,8 @@
 # Original: #"project_reader_api": "role:reader and project_id:%(project_id)s"
 # replace due to keystone app credential not honoring implicit roles
 "project_reader_api": "(role:reader or role:member) and project_id:%(project_id)s"
+
+# permit flavors with disk=0 to be used by non-admin users
+# used to skip nova size check for baremetal flavor.
+# Default is admin-only.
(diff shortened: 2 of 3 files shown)


msherman64:
Had some sanity checks via slack, persisting here for reference:

We have the following "goals" for config related to disk image size:

1. feature: allow users to launch using "big enough" images for their data
2. feature(ux): if an image won't fit on a given node, give users a good error message at API request time, not having to wait until the deploy fails
3. feature(reliability): prevent users from breaking the ironic-conductor by causing a giant image to be converted from qcow2 -> raw, which will use up cpu cycles and disk space on the conductor.
4. feature(quota/usage): prevent users from filling glance up
5. feature(snapshot): cc-snapshot should work no matter how large an allowable image is used.

the knobs we have available are:
* the flavor disk size
* glance config: [largest allowed image ](https://docs.openstack.org/glance/2023.1/configuration/glance_api.html#DEFAULT.image_size_cap)image_size_cap
* keystone quota for "glance storage per project"
* ironic configs: disk provisioning
  * [default.force_raw_images](https://docs.openstack.org/ironic/latest/configuration/config.html#DEFAULT.force_raw_images)
  * [agent.stream_raw_images](https://docs.openstack.org/ironic/latest/configuration/config.html#agent.stream_raw_images)
  
  
 
 Right now, the default 20gb disk size mostly prevents having to care about goals 2 and 3.
 We probably need to address other config related to ironic image conversion before "uncapping" the images safely.
 
 The smallest disk currently in the reference repo is 120Gb. we could set a fixed image size somewhere between 40gb (needed for biggest cuda so far) and 100gb (some margin). we would want to check if these large images can be successfully snapshotted.

msherman64:
Set back to draft in favor of: https://github.com/ChameleonCloud/chi-in-a-box/pull/403

msherman64:
For clarity, the biggest reason not to merge this is the potential for a user to request a 1tb image to get converted on the ironic conductor.

PR #403, opened 2026-05-26
feat: bump ironic flavor size to 100gb

msherman64:
The smallest disk in referenceapi is 120GB. Bump the flavor size from 20GB to 100GB to allow larger images to be used. (we can argue about what the size should be).

### DIFF
--- a/roles/post_ironic/defaults/main.yml
+++ b/roles/post_ironic/defaults/main.yml
@@ -12,7 +12,7 @@ baremetal_flavor:
   name: baremetal
   is_public: yes
   # set to useful value, must be smaller than smallest node disk
-  disk: 20
+  disk: 100
   ram: 1
   vcpus: 1
   extra_specs: 


Mark-Powers (approved):
This seems fine. Can we make this configurable so each site can adjust according to their hardware?

Can you clarify the issues in the approach in https://github.com/ChameleonCloud/chi-in-a-box/pull/399 where we set disk to 0?

It should be up to the user to ensure their disk image fits on a node. We already have some images that don't work on all nodes.

msherman64:
> This seems fine. Can we make this configurable so each site can adjust according to their hardware?
> 
> Can you clarify the issues in the approach in #399 where we set disk to 0?
> 
> It should be up to the user to ensure their disk image fits on a node. We already have some images that don't work on all nodes.

This is configurable in that I moved the config from the task to defaults, so the dict is overridable, but I could make the disk size in particular configurable via a standalone var.

With the explicit size set, I'm hesitant to make it different across sites, as that makes it difficult to reason about "will image X work at site Y", and we hide the baremetal flavor specs from users.

We could always create extra, per-project flavors with different limits as an escape hatch for power users.

PR #405, opened 2026-06-02
port chi@edge to 2023.1

msherman64:


kolla/defaults.yml
```diff
@@ -93,6 +93,17 @@ enable_cinder: no
 
 # Doni
 enable_doni: yes
+default_balena_fleet: "chi-edge-workers"
+doni_balena_fleet_mapping:
```
Mark-Powers, line 97: Can these be easily overridden per device type? Seems like a key-value dictionary might be easier, but not a huge deal.

Mark-Powers (approved):

PR #419, opened 2026-09-17
Disable blazar v2 API

Mark-Powers:
Mitigates authorization bypass in https://bugs.launchpad.net/ossa/+bug/2162719

This API is not used by us, and so can be safely disabled.


msherman64 (approved):

pdmars (approved):

codyhammock (approved):
no objection.

